PR #20, opened 2026-01-13
feat(voice/clone): support MP3 and other audio formats for reference voices

Und3rf10w:
## Summary
- Adds support for MP3, FLAC, OGG, and other audio formats when uploading reference voices for cloning
- Replace WAV-only `audio_read()` with a `soundfile`-based implementation
- Update web UI file picker to accept multiple audio formats

## Changes
- `pocket_tts/data/audio.py`: Use `soundfile` library instead of Python's `wave` module
- `pocket_tts/main.py`: Preserve original file extension in temp files for format detection
- `pocket_tts/main.py`: Add missing `truncate=True` to CLI generate command
- `pocket_tts/static/index.html`: Update file input to accept `.wav,.mp3,.flac,.ogg,.m4a,audio/*`
- `pyproject.toml`: Add `soundfile>=0.12.0` dependency

## Tested
- [x] Upload a WAV file as reference voice and generate speech
- [x] Upload an MP3 file as reference voice and generate speech
- [x] Upload a FLAC file as reference voice and generate speech
- [x] Verify CLI `--voice` option works with MP3 files

### DIFF
--- a/pocket_tts/main.py
+++ b/pocket_tts/main.py
@@ -236,7 +236,7 @@ def generate(
         )
         tts_model.to(device)
 
-        model_state_for_voice = tts_model.get_state_for_audio_prompt(voice, truncate=True)
+        model_state_for_voice = tts_model.get_state_for_audio_prompt(voice)
         # Stream audio generation directly to file or stdout
         audio_chunks = tts_model.generate_audio_stream(
             model_state=model_state_for_voice,


gabrieldemarmiesse:
Thank you for the pull request! It's a cool feature :) I'd like to keep the amount of dependencies to a mimimum if possible. Let's try to make soundfile an optional dependency.

jarodise:
I added a pre-processing layer with ffmpeg to tackle this problem.

Und3rf10w:
> Thank you for the pull request! It's a cool feature :) I'd like to keep the amount of dependencies to a mimimum if possible. Let's try to make soundfile an optional dependency.

Addressed in [252ab03](https://github.com/kyutai-labs/pocket-tts/pull/20/commits/252ab031204df2e92979a6b59b96ccc984744774)

teddybear082:
my two cents: ffmpeg can have a lot of issues so make sure if someone just wants to use the original .wav approach nothing will fail if the user doesnt have ffmpeg

vvolhejn (approved):
Thank you!

PR #21, opened 2026-01-13
Add offline mode support for HuggingFace Hub downloads

Und3rf10w:
## Summary
- Add support for running in offline mode after models have been cached
- Check `HF_HUB_OFFLINE`, `TRANSFORMERS_OFFLINE`, and `HF_DATASETS_OFFLINE` environment variables
- Pass `local_files_only=True` to `hf_hub_download` when offline mode is enabled

Basically, if `HF_HUB_OFFLINE=1` is set, you can now run `pocket-tts` if the models are cached, allowing `pocket-tts` to be ran offline

### DIFF
--- a/pocket_tts/utils/utils.py
+++ b/pocket_tts/utils/utils.py
@@ -1,5 +1,6 @@
 import hashlib
 import logging
+import os
 import time
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
Hi @Und3rf10w thanks for the pull request! Thanks for noticing the offline issue, I found another way to fix it that doesn't require the user to do anything :) https://github.com/kyutai-labs/pocket-tts/pull/27

PR #25, opened 2026-01-14
Fix apostrophe handling and server reload 500 error

jarodise:
- Fix: Normalize smart quotes (e.g., ’) to straight quotes (') in `prepare_text_prompt` to prevent tokenization issues and mispronunciation of contractions.
- Fix: When running `serve` without reload, pass the `web_app` instance directly to `uvicorn` to preserve the initialized global `tts_model`, resolving 500 errors caused by `NoneType` model state.
- Add: Warning log when using `--reload` noting potential state initialization issues.


vvolhejn:
Interesting, I can reproduce the error with `--reload` that you're seeing. @gabrieldemarmiesse did `--reload` actually work for you? If not, let's just remove the option altogether. If it works for Gabriel on Windows, it might be a Windows vs MacOS thing (not sure why that'd be the case though?) and then we can keep the option with a warning.

For now could you also add a warning to the help string of `--reload` saying that it might not work?

gabrieldemarmiesse:
I never used the - - reload option on my end

vvolhejn:
Ok, then let's just remove the `--reload` option altogether and add a comment above the uvicorn start line that says why `reload=True` doesn't work.

PR #29, opened 2026-01-15
Add support for dev environment using Nix flakes

crertel:
This should make it a lot easier for folks using Nix and NixOS to quickly setup a dev environment and run this project.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,99 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1768305791,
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
Thank you for the pull request! Since no one in the team is using NixOs, we don't want to support it in our codebase. What we can do though is add a link in the readme to your fork saying something like "If you run NixOS, you can head to ... which has the flake files making the setup easier". Would that be okay for you?

crertel:
That'd be fine! Let me set that up on my end. :)

PR #32, opened 2026-01-15
Slice the kvcache when getting the model state for a given voice

Arihant25:
#19

PR #36, opened 2026-01-15
fix: close temp file handle before deletion on Windows

wooxus:
On Windows, files cannot be deleted while a handle is still open. The previous code called os.unlink() inside the `with` block, where the NamedTemporaryFile handle was still open, causing:

> PermissionError: [WinError 32] The process cannot access the file because it is being used by another process

This fix moves the try/finally block outside the `with` statement, ensuring the file handle is closed before attempting to read and delete the file. This makes the code work correctly on Windows while maintaining the same behavior on Linux/macOS.

`uvx pre-commit run --all-files` passing

image

### DIFF
--- a/pocket_tts/main.py
+++ b/pocket_tts/main.py
@@ -149,14 +149,15 @@ def text_to_speech(
         with tempfile.NamedTemporaryFile(delete=False, suffix=".wav") as temp_file:
             content = voice_wav.file.read()
             temp_file.write(content)
-            temp_file.flush()
+            temp_file_path = temp_file.name
+        # File is now closed after exiting the `with` block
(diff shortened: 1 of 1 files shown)


wooxus:
just realised there is a pr for this already https://github.com/kyutai-labs/pocket-tts/pull/32. closing this one

gabrieldemarmiesse:
No worries, I think I'll handle this one myself as I have another solution in mind

wooxus:
> No worries, I think I'll handle this one myself as I have another solution in mind

great, thanks! I will watch the space for the fix. In the meantime I'm using this workaround.

PR #40, opened 2026-01-16
Added Colab usage and code with no need of HF Authentication

HarishDevLab:
This PR adds documentation and example code for using local TTS voices
without Hugging Face authentication.

It also includes a public Google
Colab notebook to simplify setup and testing.

No functional model changes are introduced.


vvolhejn:
Thank you, good idea! Closing in favor of #67

Issue #41, opened 2026-01-16
Audio_file conversion to Wav


staru09:
Although the project expects users to upload a WAV file for voice cloning, many users won’t have their audio in WAV format, and no helper function currently exists to convert other audio formats into the required WAV format.

While the code does perform some preprocessing such as resampling to 24 kHz and converting the audio to mono, I think it would be useful to have a helper function to automatically convert the audio in required format.

I think using FFMPEG for this conversion would be a good way to handle it.

vvolhejn:
Working on it in #20

staru09:
Cool thing
So sorry I didn't checked that PR before.

PR #42, opened 2026-01-16
added ffmpeg helper for audio type-conversion

staru09:
Related to #41


vvolhejn:
Hi, thanks for the PR. However this is a duplicate of #20 and we are planning to merge that one because it's already been reviewed and it only has `soundfile` as a dependency. We'll merge #20 in soon.

PR #44, opened 2026-01-16
Docker package build and fixed host and port

yadirhb:
These are all the fixes I made to have GH building and publishing the container image for faster usage, like:

```
docker run -p 8000:8000 --rm ghcr.io/kyutai-labs/pocket-tts
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM ghcr.io/astral-sh/uv:debian
 
 WORKDIR /app
+
 COPY ./pyproject.toml .
 COPY ./uv.lock .
 COPY ./README.md .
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,48 @@
+name: Build and Push Docker Image
+
+on:
+  push:
+    branches: [ main ]
+  workflow_dispatch:
```
gabrieldemarmiesse: Maybe we should trigger it when we do a release instead? This is what we do for the PyPI upload. I'll let you take a look at the other workflow file. It would be cool if the docker images could look like this 

```
ghcr.io/kyutai-labs/pocket-tts:v1.0.1
```
so that way we can quickly try multiple versions. I think you can do this by grabbing the name of the ref being used to run the workflow.

gabrieldemarmiesse:
Closing as stale. Feel free to re-open it if needed!

PR #58, opened 2026-01-17
fix: respect frames_after_eos parameter in generate_audio_stream()

batteryshark:
The `frames_after_eos` parameter in `generate_audio_stream()` and `generate_audio()` was accepted but never used - the code always used the auto-calculated guess.

**Before:** User passes `frames_after_eos=0` → ignored, uses guess (3+ frames of silence)
**After:** User passes `frames_after_eos=0` → respected, minimal trailing silence

Default behavior preserved when `frames_after_eos=None`.

I *think* this is the intention of that argument.


vvolhejn:
Ah, good catch! Thank you.

Issue #57, opened 2026-01-17
Any way to run from model files saved in custom location?


teddybear082:
Hi, I signed the hugging face user agreement and downloaded the model files locally from the hugging face page to enable voice cloning but I can’t find a way to point the TTSModel to those local files on init; it seems entirely hard coded to download from hugging face at runtime or use hugging face cache. Am I missing some parameter I can use when initiating the model to point to a folder (eg C://pocket-tts/) where I have the model files installed?

staru09:
I think you might want to check hf cache because this is where weights are saved when downloaded from hf
https://huggingface.co/docs/huggingface_hub/en/guides/manage-cache

gabrieldemarmiesse:
It's not possible yet but it's a very valid feature request :)

teddybear082:
> It's not possible yet but it's a very valid feature request :)

Ok thank you! from reviewing the files to try to ensure I was not missing something before posting here I think I saw the points of entry for this so if someone else does not get to it first I will try to do a PR to support this for next weekend.

teddybear082:
Closing issue since PR was accepted for this and will be in next version.

PR #59, opened 2026-01-18
Basic changes to allow loading of local model files

teddybear082:
-modify main,py to not hardcode using DEFAULT_VARIANT

-allow submission of path string to path of variant.yaml file saved in custom directory in lieu of model variant identifier in CLI to use when loading model

-addresses #57

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,10 +110,6 @@ you have multiple voices you want to use. `load_model()`
 and `get_state_for_audio_prompt()` are relatively slow operations,
 so we recommend to keep the model and voice states in memory if you can.
 
-As above, you can also load a pocket-tts model using custom paths in a locally-saved .yaml file by passing
-the path to the .yaml file as an argument for TTSModel.loadmodel:
-`tts_model = TTSModel.load_model("path/to/config.yaml")`
-
 You can check out the [Python API documentation](https://github.com/kyutai-labs/pocket-tts/tree/main/docs/python-api.md) for more details and examples.
 
 ## Unsupported features


teddybear082:
I'm not sure what is failing still on the pre-commit check. EDIT: it was literally just spaces ha.

vvolhejn:
Hi! Thanks for the PR. This does the same thing as #61 , though I like this approach better.

Could you elaborate on what your use case is here? Why do you need to change the path at all? If you need Hugging Face to download/look for models in a different path, you can configure HF_HOME, see [here](https://huggingface.co/docs/datasets/en/cache)

If that doesn't work, why not just copy the config and change the paths manually in the YAML? Do we need code specifically for this?

teddybear082:
Thank you for considering the PR.

Philosophically, two benefits of running local models are not needing internet access and feeling like you have control over what is on your device.  By allowing the user control over their own download, if they want to, it obviates what may feel like a shadow download to a shadow location to the user happening without their awareness.  Not as important for front line devs but for people who create frameworks, it’s more palatable to say “the program will download the model for you but if you don’t trust that, or don’t want it to, just download it yourself, place it in a directory you want (eg maybe they have a folder where they place all local models) and set the command line argument to the model (or more likely for this type of user, put the path to your model in a UI designed by the dev).

Practically, this type of feature also solves for later fine tuned models that have been fine tuned locally by the user or different versions of the model to be used over time versus whatever might be latest on hugging face, which can help stability and prevent things from breaking. For instance, chatterbox has been evolving their models, I made a framework that worked with one version of it, since I have the model on my computer and can direct others to the same model I used and have them download it, my framework is stable and I can choose to update it if I want, but don’t need to. If the only option was the code fetching the latest model from hugging face my code might keep breaking or I depend on the devs of the model to always ensure backwards compatibility which they may or may not do.

thank you!

vvolhejn:
Thanks for elaborating! I understand the need better now. But wouldn't it be cleaner to just copy the existing config and changing the paths in the YAML? Maybe we can replace the `--variant` flag with a `--config` that will either load from the standard config location if the argument doesn't end with `.yaml`, and look for a config file in the path that you specify if it does?

That way seems less magical to me, the paths specified in the config will really be the ones that end up being loaded. The disadvantage is a bit more boilerplate because you'd be copying the model architecture configuration, but that doesn't seem too bad.

teddybear082:
Yes that makes sense, I can implement it that way, I think the code would be basically the same except instead of doing it through the variant argument, the alternative process would be triggered by the new —config argument.  I’m not sure if it’s as simple as loading wherever paths are in the custom yaml though because right now I’m pretty sure the code uses hugging face to load those and it may not accept local path as as an argument but I’m sure we can just skip that if a local file path is detected. I will try and see.

vvolhejn:
Weights are loaded using the [`download_if_necessary()`](https://github.com/kyutai-labs/pocket-tts/blob/bfbaf80437936eed2c1063eb5099733160f15ce9/pocket_tts/utils/utils.py#L68) function, which will just let local paths through unchanged. Should be fine!

teddybear082:
> Weights are loaded using the [`download_if_necessary()`](https://github.com/kyutai-labs/pocket-tts/blob/bfbaf80437936eed2c1063eb5099733160f15ce9/pocket_tts/utils/utils.py#L68) function, which will just let local paths through unchanged. Should be fine!

Ok thank you for the input, this seems to work fine as you noted.

PR #60, opened 2026-01-18
fix: close temp file before reading on Windows

aoc81:
## Problem
On Windows, the \/tts\ endpoint fails with a \PermissionError\ when uploading a voice file for cloning:

\\\
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process
\\\

## Cause
\NamedTemporaryFile\ keeps the file open in the context manager. On Windows, files cannot be deleted (or read by another process) while still open.

## Fix
- Close the temp file explicitly before \get_state_for_audio_prompt()\ reads it
- Catch \PermissionError\ on cleanup in case the file is still in use

Tested on Windows 10.

### DIFF
--- a/pocket_tts/main.py
+++ b/pocket_tts/main.py
@@ -146,17 +146,22 @@ def text_to_speech(
         logging.warning("Using voice from URL: %s", voice_url)
     elif voice_wav is not None:
         # Use uploaded voice file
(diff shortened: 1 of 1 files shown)


vvolhejn:
Sorry, I missed this one! I fixed it in #72 . Could you try if that works for you?

aoc81:
All good, thanks!

PR #61, opened 2026-01-18
feat: add support for local model path via POCKET_TTS_LOCAL_MODELS_PATH

### LINKED ISSUE #57: Any way to run from model files saved in custom location?

aipdv:
This PR addresses Issue #57 

- It introduces the `POCKET_TTS_LOCAL_MODELS_PATH` environment variable. 
- If set, the system checks this local directory for model files (either in a flat structure or mirroring the Hugging Face path) before attempting to download them.
- Validated with a new test case `tests/test_custom_path.py`

running of test case

Closes #57


gabrieldemarmiesse:
I believe it would make more sense to have the files paths as a cli argument and argument to the load_model function. Does it have to be an environment variable?

aipdv:
That makes sense. I'll refactor this to add a --model-path CLI argument and a model_path parameter to load_model(). Environment variables were my first instinct for caching scenarios, but you're right that explicit arguments are more discoverable and Pythonic.

I'll push an update soon.

aipdv:
Refactored the implementation based on your feedback! 

Changes:
- Removed the `POCKET_TTS_LOCAL_MODEL_PATH` environment variable
- Added `model_path` parameter to `TTSModel.load_model()` 
- Added `--model-path` CLI flag to both `generate` and `serve` commands
- Updated `download_if_necessary()` to accept the path explicitly

The API is now much cleaner and more discoverable. Users can simply run:
```bash
pocket-tts generate --model-path ./my-models --text "Hello world"
```

All tests and pre-commit hooks are passing.

vvolhejn:
Hi! Thanks for the PR. I think we can do this in a simpler way. Notice that [here](https://github.com/kyutai-labs/pocket-tts/blob/36f21a0a7f9356f31691ad7a3a43e9cfbfba44ad/pocket_tts/models/tts_model.py#L200) we load a config and the config contains paths to the weights and the tokenizer. If you need a local model, why not simply change the config?

Could you elaborate on what your use case is here? Why do you need to change the path at all? If you need Hugging Face to download/look for models in a different path, you can configure `HF_HOME`, see [here](https://huggingface.co/docs/datasets/en/cache)

aipdv:
Hi @vvolhejn , thanks for the review!

I investigated `HF_HOME` as an alternative, but it doesn't support the "flat directory" use case described in issue #57 . 

I ran a reproduction script setting `HF_HOME` to a local folder containing the model files. As shown in the ss below, `huggingface_hub` fails to detect the local files because it strictly enforces the internal `snapshots/<hash>/` directory structure:

screenshot from cli 
---
code screenshot

vvolhejn:
I'll continue the discussion in #59

PR #31, opened 2026-01-15
Add docker-compose.yml and example.env

schkng:
PR related to #28

PR #64, opened 2026-01-18
feat: add docker-compose.yaml with model cache persistence 

### LINKED ISSUE #28: Introduce docker-compose.yaml
I'm interested in running pocket-tts containerized, I run about 20 containers on my home server already and it would fit nicely in. I can see a Dockerfile in the repo already, having a docker-compose file would make the following more visible to the user:

* any available environment variables
* any available ports (guess there's just the one)
* a volume/bindable folder to download the models to so that they can be reused the next time

As all this would be configurable in the file. Simple `docker container run` re-downloads the model data on every re-run which is wasteful and I don't believe you use your docker image like this.

There's one more reason why I'd like this. Running the `uvx pocket-tts serve` will at the moment, despite the CPU variant of torch being mentioned literally everywhere in pyproject.toml and uv.lock, ignore the CPU variant request and instead download multiple gigabytes of torch CUDA variant on my Linux that has no Nvidia GPU.

aipdv:
Closes #28

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,18 +1,12 @@
 services:
   pocket-tts:
     build: .
-    container_name: pocket-tts
     ports:
       - "8000:8000"
     volumes:
-      # Persist model cache across container restarts to avoid re-downloading
       - pocket_tts_cache:/root/.cache/pocket_tts
       - hf_cache:/root/.cache/huggingface
-    environment:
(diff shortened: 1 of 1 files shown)


aipdv:
This PR provides a simplified `docker-compose.yaml`
(addressing feedback in #31 and #28)

It is designed to be zero-config (no .env required) while maintaining model persistence across restarts via named volumes.

Closes #28

PR #68, opened 2026-01-19
fix: support multi-channel audio prompts in audio_read

aipdv:
Currently, audio_read assumes all input WAV files are mono. When a user provides a stereo recording, the function flattens the interleaved PCM data into a single dimension. 

This results in a 'broken' voice clone where the pitch is shifted down by an octave and the speech tempo is halved, as the model interprets two channels of data as a single mono stream of double the duration.

Changes:
* Updated audio_read to check n_channels via the wave module.
* Implemented specific reshaping for multi-channel inputs.
* Added a downmix step that averages channels to mono before passing the tensor to the conditioning encoder.

This fix ensures that standard stereo captures from mobile devices or external mics work out of the box for zero-shot cloning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,10 @@ Forget about the hassle of using GPUs and web APIs serving TTS models. With Kyut
 
 Supports Python 3.10, 3.11, 3.12, 3.13 and 3.14. Requires PyTorch 2.5+. Does not require the gpu version of PyTorch.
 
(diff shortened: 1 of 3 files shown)


vvolhejn:
Hi! We've recently merged #20, which should hopefully fix the issue for you. Could you let us know if it works for stereo inputs on latest `main`?

aipdv:
thanks for pointing it out, it works now!!

PR #81, opened 2026-01-24
Fix voice cloning from CLI

erikos:
The CLI was passing voice as a string instead of a Path object, and was missing truncate=True. This matches the working WebUI implementation.

### DIFF
--- a/pocket_tts/main.py
+++ b/pocket_tts/main.py
@@ -240,7 +240,7 @@ def generate(
         )
         tts_model.to(device)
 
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
I think what we'll want to do instead is enable the truncate only when the user asks for it and raise an error when the audio is too long. We don't want users thinking that it's useful to pass one hour of audio as input to the model if we truncate anyway. The server is used for quick experimentation so I believe we'll still truncate silently there.

PR #89, opened 2026-01-29
Add safetensors support to the server

thataboy:
Allow using custom voices in server requests.

- Create new `pocket_tts.default_parameters `CUSTOM_VOICES_DIR` (default: `<repo-root>/voices`)
- Audio files added to CUSTOM_VOICES_DIR will be automatically converted to safetensors on server start. The voice can then be used by passing it by name name via `voice_url`, just as if it's one of the `PREDEFINED_VOICES`. (Note: Perhaps this should be renamed simply to `voice`?)
- Clients can get a list of available voices via GET '/voices'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,10 @@ test_en_audio_4000.model
 profile_output*
 *.safetensors
 *.wav
+*.mp3
+*.flac
+*.ogg
+*.aiff
 *.model
 *.pt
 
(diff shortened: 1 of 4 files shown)


gabrieldemarmiesse:
Closing to tidy up our issues and PRs but feel free to continue the discussion here, my notifications are enabled

thataboy:
Honestly, it's too small a feature to warrant its own repo and maintenance. I just thought it'd be useful to have in the official release  since this is something I've seen standard with other TTS.  But we're able to pass a safetensors into get_state_for_audio_prompt now so it's just a convenience feature and I understand you want the code to be an example rather than production ready.

PR #107, opened 2026-02-05
Add OpenAI-compatible /v1/audio/speech endpoint

oldeucryptoboi:
## Summary
- Add `POST /v1/audio/speech` as a drop-in replacement for the OpenAI TTS API
- Map OpenAI voice names (alloy, echo, etc.) to Pocket TTS voices
- Support `wav` (default) and `pcm` response formats
- Document the endpoint with usage examples in docs/serve.md

## Test plan
- [x] 8 new tests in `tests/test_openai_endpoint.py` — all passing
- [x] Full test suite passes (24/25, 1 pre-existing HF auth failure)
- [x] Linting passes
- [x] Manual smoke test with OpenAI Python client

### DIFF
--- a/docs/serve.md
+++ b/docs/serve.md
@@ -53,6 +53,58 @@ Then, use the --config option to point to your newly created config.
 pocket-tts serve --config "C://pocket-tts/my_config.yaml"
 ```
 
+## OpenAI-Compatible Endpoint
+
+The server exposes `POST /v1/audio/speech`, a drop-in replacement for the
+[OpenAI TTS API](https://platform.openai.com/docs/api-reference/audio/createSpeech).
+Existing apps only need to change their `base_url` to point at Pocket TTS.
+
+### curl example
+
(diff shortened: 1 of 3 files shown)


teddybear082:
Hi OP. I have an openai compatible server here: https://github.com/teddybear082/pocket-tts-openai_streaming_server, you may want to look at it. I think it's important you support the formats that openai supports to the fullest extent possible as well as streaming and non-streaming implementations for this to be a real "drop in replacement".

gabrieldemarmiesse:
Hi @oldeucryptoboi thanks a lot for the pull request! As @teddybear082 supporting OpenAI compatible endpoints is a significant amount of work. We prefer deferring this work to the community and focus on providing a reference implementation that's easy to run and read. Feel free to look at community projects like https://github.com/teddybear082/pocket-tts-openai_streaming_server :)

Issue #76, opened 2026-01-22
Add text normalization


vvolhejn:
For cases like #73, we should pre-process the text (apply text normalization) to make the job easier for the text to speech.

See e.g. https://developer.nvidia.com/blog/text-normalization-and-inverse-text-normalization-with-nvidia-nemo

Complicated examples:
- 2/3 -> two thirds
- $100 -> hundred dollars (or "one hundred dollars")
- 50g -> fifty grams
- Sunset Blvd -> Sunset Boulevard

The model might learn _some_ of these from the training data, but we can make its job easier.

sanjaymarison:
Different English version pronounce the numbers differently, like some say 1 lakh dollars but americans might say hundred thousand dollars. So maybe account for that as well?

teddybear082:
This may help (MIT licensed): https://github.com/keithito/tacotron/blob/master/text/cleaners.py

PR #111, opened 2026-02-08
Add some text normalization. convert unicode to ASCII and expand numbers into words.

handpickencounter:
Add some text  normalization.
Improves overall TTS quality.

https://github.com/kyutai-labs/pocket-tts/issues/76

todo: split_into_best_sentences() can split in the middle of quoted text (characters having conversation in a book) which can ruin a flow.

partial workaround: increase max number of tokens from 50 to 100 or more.

### DIFF
--- a/pocket_tts/models/tts_model.py
+++ b/pocket_tts/models/tts_model.py
@@ -789,7 +789,7 @@ class TTSModel(nn.Module):
 
 def wordify_numbers(text):
     def replace_num(match):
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
Hello, I believe we'll want to encourage people to use an external text normalization library, otherwise we might end up with a lot more code inside pocket TTS than what we can maintain.

Thanks for the PR!

PR #119, opened 2026-02-11
ADDED MUTLICORE SUPPORT

TheJoshCode:
i added a num_threads arg to load_model in your tts_model.py.

Tested this and it's working great in my Booklet app, I use a mutli-worker setup to get 5x the single core speed since multiple model insts are loaded.

[https://github.com/TheJoshCode/Booklet](https://github.com/TheJoshCode/Booklet)


pocket_tts/models/tts_model.py
```diff
@@ -213,6 +215,10 @@ def load_model(
                 are not found.
             ValueError: If the configuration is invalid or incompatible.
         """
+        if num_threads is not None:
+            torch.set_num_threads(num_threads)
```
gabrieldemarmiesse: Wouldn't it be easier to ask users to call `torch.set_num_threads(num_threads)` after importing pocket TTS? Since this is setting a global variable, there is no need to do it in a function. We could just add to a doc file

```python
import pocket_tts
torch.set_num_threads(4)
```
for example

gabrieldemarmiesse (commented):
Thanks for the PR!

pocket_tts/models/tts_model.py
gabrieldemarmiesse: Maybe we can add it to the `generate` and `serve` command. That would be useful no? People can then play with it to find what's the best speed. `--num-threads` and we would have `torch.set_num_threads(num_threads // 2)` since we always have one thread for CALM and one thread for the mimi decoding.

danneauxs: I've independently implemented this. I use four cores to six cores of my 8 on my CPU.    I usually stick with four because increasing past that gives gains but they're proportionately smaller gains.   Using 4 cores I usually get 6min of audio for 1 min processing or 6x+.   Add in some batching and you can get that up to about 10 times or in other parlance one tenth real time.

gabrieldemarmiesse:
Closing as stale, feel free to reopen or make a new PR if you want to continue this work!

PR #137, opened 2026-02-25
Raise minimum huggingface_hub to 0.13.0 for consistent offline behavior

joshwhiton:
## Summary
Raise the minimum `huggingface_hub` version from `>=0.10` to `>=0.13.0`.

## Motivation
`TRANSFORMERS_OFFLINE` behavior in `hf_hub_download(...)` varies across the currently declared range:
- `0.12.1`: `TRANSFORMERS_OFFLINE=1` is not enforced in this call path.
- `0.13.0`: `TRANSFORMERS_OFFLINE=1` is enforced (`LocalEntryNotFoundError` on empty cache).

This PR tightens the dependency floor to the first version where this works out of the box.

## Scope
- Only dependency floor bump in `pyproject.toml`.
- No code-path changes in `download_if_necessary()`.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1152,7 +1152,7 @@ requires-dist = [
     { name = "beartype", specifier = ">=0.22.5" },
     { name = "einops", specifier = ">=0.4.0" },
     { name = "fastapi", specifier = ">=0.100" },
-    { name = "huggingface-hub", specifier = ">=0.10" },
+    { name = "huggingface-hub", specifier = ">=0.13.0" },
     { name = "numpy", specifier = ">=2" },
     { name = "pydantic", specifier = ">=2" },
     { name = "python-multipart", specifier = ">=0.0.21" },


gabrieldemarmiesse:
Can you give me a reproducible example for the bug you're trying to fix? Currently pocket-tts works offline, here is an example on my laptop:

```
$ ping www.google.com -W 5
ping: www.google.com: Temporary failure in name resolution

$ uvx pocket-tts generate --voice /tmp/voice.wav
INFO: Loading sentencepiece tokenizer from hf://kyutai/pocket-tts-without-voice-cloning/tokenizer.model@d4fdd22ae8c8e1cb3634e150ebeff1dab2d16df3
INFO: Loading TTSModel weights from hf://kyutai/pocket-tts/tts_b6369a24.safetensors@427e3d61b276ed69fdd03de0d185fa8a8d97fc5b
INFO: Encoding audio prompt took 1051 ms
INFO: Prompting audio took 218 ms
INFO: Size of the model state for audio prompt: 3 MB
INFO: starting timer now!
INFO: Prompting text took 120 ms
INFO: Average generation step time: 23 ms
INFO: Waiting for mimi decoder to finish took 0 ms
INFO: Generated: 6080 ms of audio in 1945 ms so 3.13x faster than real-time
INFO: Results written in ./tts_output.wav
INFO: --------------------
INFO: If you want to try multiple voices and prompts quickly, try the `serve` command.
INFO: If you like Kyutai projects, comment, like, subscribe at https://x.com/kyutai_labs
```

joshwhiton:
Thanks for checking. I reproduced your working result too.

However, when testing across the declared huggingface_hub>=0.10 range: HF hub 0.10.1 did **not** enforce TRANSFORMERS_OFFLINE (while HF_HUB_OFFLINE did).

In other words, whether TRANSFORMERS_OFFLINE is enforced varies across Pocket’s allowed dependency range (huggingface_hub>=0.10).

Since TRANSFORMERS_OFFLINE is a user-facing offline signal many users expect to work, explicitly mapping offline env vars to local_files_only=True makes Pocket deterministic, independent of hub-version quirks.

Happy to close this PR as unnecessary for modern hub versions, or retarget it explicitly as legacy behavior normalization across your declared HF hub range.

gabrieldemarmiesse:
Maybe let's increase the minimum version of huggingface then, until we have a minimum version where the offline mode works out of the box. It will be much simpler to maintain

gabrieldemarmiesse:
Thanks!

joshwhiton:
Sounds good. Identified huggingface 0.13.0 as the first version where TRANSFORMERS_OFFLINE=1 is properly enforced. This PR now only tightens the dependency floor to that version.

Sorry for raising such a minor issue, but my app prides itself on being fully offline. And it may have been more pronounced when I first encountered this while implementing the very first version of Pocket.

PR #140, opened 2026-03-03
Add project to README.md

tonelord:
Add pocket-tts-deno to project list. It's an AI-assisted port of [pocket-tts-server](https://github.com/ai-joe-git/pocket-tts-server) to deno. I wanted something like a simple small decent quality TTS server, so this made a lot of sense to me.


gabrieldemarmiesse:
Thanks!

PR #142, opened 2026-03-10
Add PocketTTS.cpp to alternative implementations

VolgaGerm:
Adds PocketTTS.cpp to the alternative implementations list — a C++ ONNX Runtime port with CLI, streaming HTTP server, and a C API for FFI.


gabrieldemarmiesse:
That's pretty cool! Nice project!

PR #143, opened 2026-03-12
Split long sentences on commas to prevent skipped words

costajohnt:
## Summary

When a single sentence exceeds `max_tokens` (50) and has no sentence-ending punctuation (`.`, `!`, `?`), the text splitting logic now falls back to splitting on commas, semicolons, and colons. This prevents the model from silently skipping parts of long sentences.

This implements **Path 2** from [the maintainer's analysis](https://github.com/kyutai-labs/pocket-tts/issues/38#issuecomment-3913789520):

> When we encounter very very long sentences like this one, and thus cannot split, we split on other characters, like a "," for example.

**Changes:**
- Extract boundary-finding logic into reusable `_find_boundary_indices` and `_segments_from_boundaries` helpers
- Add comma/semicolon/colon fallback splitting for oversized sentence segments
- Log a warning when a chunk still exceeds `max_tokens` after all splitting attempts
- Add 8 unit tests covering the splitting behavior

Fixes #38

PR #141, opened 2026-03-05
added warning for long sentences

RunjiaChen:
Hello, I attempted to add a warning when the sentences are extremely long, which may result in some phrases being left out. This is to address the issue mentioned in https://github.com/kyutai-labs/pocket-tts/issues/38. 

This is my first time contributing to a open sourced library, I appreciate any feedback regarding how I can improve. Thank you!

PR #148, opened 2026-03-20
Lower default chunk size to fix missing words in generation

### LINKED ISSUE #38: Generation misses parts of sentences
(Sorry, posted on HN first, but posting here for tracking) 

So, on my M1 mac, did `uvx pocket-tts serve`. Plugged in

> It was the best of times, it was the worst of times, it was the age of wisdom, it was the age of foolishness, it was the epoch of belief, it was the epoch of incredulity, it was the season of Light, it was the season of Darkness, it was the spring of hope, it was the winter of despair, we had everything before us, we had nothing before us, we were all going direct to Heaven, we were all going direct the other way—in short, the period was so far like the present period, that some of its noisiest authorities insisted on its being received, for good or for evil, in the superlative degree of comparison only

(Beginning of Tale of Two Cities) into the web interface. Using javert

but the problem is Javert skips over parts of sentences! Eg, it starts:

> "It was the best of times, it was the worst of times, it was the age of wisdom, it was the epoch of belief, it was the epoch of incredulity, it was the season of Light, it was the spring of hope, it was the winter of despair, we had everything before us, ..."

Notice how it skips over "it was the age of foolishness,", "it was the winter of despair,"

Which... Doesn't exactly inspire faith in a TTS system.

Marius seemed to do better...

Greal-dev:
Fixes #38: Generation was skipping parts of sentences due to the default `MAX_TOKEN_PER_CHUNK` being too large (50). Long sentences with no punctuation splitting points were processed as a single large chunk, causing the model to drop words. Lowering the default chunk size to 25 reduces errors in generation for long sentences.

Addresses #38.


Greal-dev:
Closing — PRs #141 and #143 already propose fixes for this issue. Apologies for the duplicate.

PR #155, opened 2026-04-01
Add french, italian, portuguese, spanish, german

gabrieldemarmiesse:
The french 6 layers variant wasn't included because it's not good enough yet. Use the `french_24l` variant for now.


benterris:
Thanks a lot for your amazing work! 

As you were requesting some feedback on the issue: I played around with the `french_24l` model, and noted that it struggled a lot with semicolons, consistently producing lots of artifacts around them. Replacing them with `.` solves the issue.

Here is an example prompt and generated output: 
```
Des figuiers entouraient les cuisines ; un bois de sycomores se prolongeait jusqu'à des masses de verdure, où des grenades resplendissaient parmi les touffes blanches des cotonniers ; des vignes, chargées de grappes, montaient dans le branchage des pins ; un champ de roses s'épanouissait sous des platanes ; de place en place sur des gazons se balançaient des lis ; un sable noir, mêlé à de la poudre de corail, parsemait les sentiers, et, au milieu, l'avenue des cyprès faisait d'un bout à l'autre comme une double colonnade d'obélisques verts.
```
Output: 
[tts_output_semicolons.zip](https://github.com/user-attachments/files/26595799/tts_output_semicolons.zip)

gabrieldemarmiesse:
Thanks for trying it out! It's exactly the kind of feedback we need! I'll add some text preprocessing for the french model to remove the semicolons :)

teddybear082:
Hi I pulled this branch and was able to load and generate german speech so far. I'm not a native speaker but shared with those who were and they agreed it sounded good in two different voices (one built in and one cloned, and the cloned voice was a sample with a german accent but speaking english).  So that's good! Looking forward to an actual Pypi release to make it easier to test in my more robust projects like my server and WingmanAI, because right now I'm just generating some sample lines in a test script.  The german model tested was the 6 layer variant not the full one.

I think I have this right but want to check - .safetensors voices are only compatible with the language model they were generated with, correct? I did a quick test and that seemed to be the case as using a .safetensors from german with the new english voice was extremely garbled, and I also noticed there are separate "embeddings" folders with the .safetensors for each voice.

gabrieldemarmiesse:
@teddybear082 indeed you're right. The embeddings aren't really embeddings, they're the kvcache of the model with the voice already prompted. So it's unique to each combo "voice - model". Thanks for testing the german model!

PR #172, opened 2026-04-27
Cancel the TTS generation if the connection to the /tts endpoint is interrupted.

memchr:
If you abort a TTS generation in the web interface (`pocket-tts serve`), for example by refreshing the page or closing it, the inference process will still run until the end of the text. This can give poor server performance if you interrupt long text generation very often.

So this change cancels the onging generation cancel the ongoing generation process if the client closes the connection prematurely.
#### Demo
unmodified server on port 8000; server with changes on port 7000.
[asciicast](https://asciinema.org/a/Kl0jHNQF1TfSWixy)

### DIFF
--- a/pocket_tts/main.py
+++ b/pocket_tts/main.py
@@ -9,12 +9,12 @@ from queue import Queue
 
 import typer
 import uvicorn
-from fastapi import FastAPI, File, Form, HTTPException, UploadFile
+from fastapi import FastAPI, File, Form, HTTPException, Request, UploadFile
 from fastapi.middleware.cors import CORSMiddleware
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
Thank you for the pull request!  The feature in itself is very useful, but the goal of the repository is to be minimalist and to show a reference implementation. The idea is that other projects can build alternative implementations because the repo is very simple, so we'll try to not increase the complexity unless absolutely necessary. Feel free to use other projects that give access to pocket-tts through a webserver, like https://github.com/teddybear082/pocket-tts-openai_streaming_server

PR #173, opened 2026-04-30
Rephrase default German sentence

vvolhejn:
"von Kyutai" is a bit easier for the TTS than "Kyutais" altough it still struggles. "du wirst mich mögen" sounded unnatural to me


gabrieldemarmiesse:
LGTM!

PR #180, opened 2026-05-04
Add default voice for each language

### LINKED ISSUE #166: Add new default voices for each new language to avoid english accents
We now have support for more languages, but all the default voices are cloned from english audio. So for example, when using the french model, people end up generating audio with an english accent when using the default voices. Users can obviously work around this for now by cloning a french voice with the french model, a german voice with the german model, etc...

What we'd like though is to provide a good first impression out of the box. We should provide at least one built-in voice in each language (french, italian, german, etc...), and when a user selects doesn't specify a voice, the default voice should correspond to the model language. So when doing 

```
uvx pocket-tts generate --language german
```
the voice used shouldn't be `alba` which is an english voice, we should select a german voice.

We encourage the community to provide good voices for diverse languages. Of course the owner of the voice should have allowed the use of the voice. We don't want to have trouble legally speaking. 

Feel free to drop a PR if you have something working!

gabrieldemarmiesse:
Fix https://github.com/kyutai-labs/pocket-tts/issues/166


docs/CLI Commands/generate.md
```diff
@@ -17,7 +17,7 @@ This will generate a WAV file `./tts_output.wav` with the default text and voice
 ### Core Options
 
 - `--text TEXT`: Text to generate (default: "Hello world! I am Kyutai Pocket TTS. I'm fast enough to run on small CPUs. I hope you'll like me.")
-- `--voice VOICE`: Path to audio conditioning file (voice to clone) (default: "hf://kyutai/tts-voices/alba-mackenna/casual.wav"). Urls and local paths are supported.
+- `--voice VOICE`: Path to audio conditioning file (voice to clone). Defaults to a built-in voice chosen from the language: `giovanni` (it) for italian, `lola` (es) for spanish, `alba` (en) otherwise. Urls and local paths are supported.
```
pinage404, line 20: If this file is true

https://github.com/kyutai-labs/pocket-tts/pull/180/changes#diff-17abf6c190e5d41647083d9089ea9975352cdfd39fc23b30e61b11399af89a4cR43

It should be

```suggestion
- `--voice VOICE`: Path to audio conditioning file (voice to clone). Defaults to a built-in voice chosen from the language: `giovanni` (it) for italian, `lola` (es) for spanish, `juergen` (de) for german, `rafael` (pt) for portuguese, `estelle` (fr) for french, `alba` (en) otherwise. Urls and local paths are supported.
```

docs/CLI Commands/generate.md
```diff
@@ -17,7 +17,7 @@ This will generate a WAV file `./tts_output.wav` with the default text and voice
...
 - `--output-path OUTPUT_PATH`: Output path for generated audio (default: "./tts_output.wav")
 - `--language LANGUAGE`: Language for the TTS model, one of `'english_2026-01'`, `'english_2026-04'`, `'english'`, `'french_24l'`, `'german_24l'`, `'portuguese_24l'`, `'italian_24l'`, `'spanish_24l'` (default: `english`, which is the same model as `'english_2026-04'`). Incompatible with `--config`. The "24l" variants are bigger models, not distilled yet and here only as preview.
```
pinage404, line 22: There are also non `24l` variants for all but `french` (sadly) https://huggingface.co/kyutai/pocket-tts-without-voice-cloning/tree/e041936c75475d350b405bc870bcf7c22da4e9e6/languages

```suggestion
- `--language LANGUAGE`: Language for the TTS model, one of `'english_2026-01'`, `'english_2026-04'`, `'english'`, `'french_24l'`, `'german'`, `'german_24l'`, `'portuguese'`, `'portuguese_24l'`, `'italian'`, `'italian_24l'`, `'spanish'`, `'spanish_24l'` (default: `english`, which is the same model as `'english_2026-04'`). Incompatible with `--config`. The "24l" variants are bigger models, not distilled yet and here only as preview.
```

gabrieldemarmiesse: Thanks for letting me know! I'll fix it!

docs/CLI Commands/generate.md
gabrieldemarmiesse: Good catch!

PR #187, opened 2026-05-09
Improve efficiency by reusing threads

mscdex:
This also improves performance a tiny bit.

### DIFF
--- a/pocket_tts/models/tts_model.py
+++ b/pocket_tts/models/tts_model.py
@@ -4,8 +4,8 @@ import math
 import os
 import queue
 import statistics
-import threading
 import time
+from concurrent.futures import ThreadPoolExecutor
 from functools import lru_cache
 from pathlib import Path
 
@@ -88,6 +88,7 @@ class TTSModel(nn.Module):
         self.pad_with_spaces_for_short_inputs: bool = pad_with_spaces_for_short_inputs
         self.model_recommended_frames_after_eos = model_recommended_frames_after_eos
         self.remove_semicolons = remove_semicolons
+        self.threads = ThreadPoolExecutor(max_workers=2)
 
     @property
     def device(self) -> torch.device:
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
Hi! Thanks for the pull request! Do you have numbers that illustrate the speedup?

mscdex:
> Do you have numbers that illustrate the speedup?

Not at the moment, but from my recollection it was maybe 1 second difference or less, but that entirely depends on how many chunks the input text ends up generating since the threads are currently (re)created for every chunk. Overall I think the primary benefit will probably be more for the garbage collector and things like that?

pixelbits-mk (approved):
**Issue:** The model now shares just 2 worker threads for every call instead of making new threads each time. If the same model is used twice at once, the second call has to wait its turn instead of running right away.

**But:** the code already says you're not supposed to run it twice at once on the same model, so this may not be a real problem in practice — depends on how you use it.

**Fix suggestion:** Either give it more workers, or make it clear/enforced that only one call runs at a time per model.

gabrieldemarmiesse:
Apologies, we won't merge performance improvements that aren't measured. If the benefit is big relative to the amount of code added, we'll merge it

manu-lm:
I tried this on current main to see whether the speedup is measurable, and to check the lifecycle. Summary: the refactor is reasonable, the performance claim is not measurable, and there are two things I would fix before merging.

**Benchmark** — 12 generations per side after warm-up, same GPU, same texts and voice, one process each:

| | median | mean | min |
|---|---|---|---|
| main | 1305.1 ms | 1282.6 ms | 956.1 ms |
| this PR | 1273.9 ms | 1257.7 ms | 929.0 ms |

That is 31 ms (2.4%) in the claimed direction, but I do not think it is real: within-side spread is ~350 ms (min 956 vs median 1305), so with n=12 a 31 ms difference is noise. The mechanism agrees — `threading.Thread` creation is roughly 80 µs, and this saves two per generation, so the ceiling on the effect is ~0.2 ms out of ~1280 ms, i.e. 0.01%. Anything larger than that in a measurement is the machine, not the change.

So I would drop the performance claim and keep this as a code-structure change, which it is: one pool instead of per-call thread construction reads better.

**Two things to fix:**

1. **The pool is never shut down.** `ThreadPoolExecutor` threads are non-daemon, while the `threading.Thread` calls this replaces were `daemon=True`. A process that constructs a `TTSModel` and returns can now hang at exit. A `close()` (plus `__enter__`/`__exit__`) calling `self.threads.shutdown(wait=True)` covers it.

2. **`self.threads.submit(run_generation)` discards the future**, so an exception inside `run_generation` is swallowed rather than raised or logged. `add_done_callback` with a small handler that calls `future.exception()` and logs it restores the previous visibility.

**It also fails CI on current main**, unrelated to the above: `tests/test_generation_regressions.py` drives `TTSModel._generate` and `generate_audio_stream` with `SimpleNamespace` fakes, which now need a `threads` attribute. Two one-line additions to the fakes fix it, but it is worth noting that the change couples these methods to model state they did not previously need.

With those three changes applied locally, all 111 tests pass.

PR #188, opened 2026-05-09
fix: rewrite decimals before splitting to avoid spurious sentence boundaries

### LINKED ISSUE #162: Pocket tokenizer incorrectly splits decimals
Main helper logic with the Pocket tokenizer incorrectly splits decimals:

The average human body temperature is 98.6°F, which is a common decimal used in medicine. -> ['The average human body temperature is 98.', '6°F, which is a common decimal used in medicine.']
Pi is 3.14 and e is 2.718. -> ['Pi is 3.', '14 and e is 2.', '718.']

Issue stems from two places:

Pocket uses raw SentencePiece tokenization with no decimal-aware normalization.
/pocket_tts/conditioners/text.py#L34-L35 , it just does self.sp.encode(text, out_type=int).

Sentence splitting still treats every . token as a sentence boundary.
/models/tts_model.py#L973-L1039), split_into_best_sentences(...) builds boundary tokens from tokenizer(".!...?") and splits on those tokens without any decimal exception.

I fixed this in my app to rewrite standalone decimals like 98.6 to 98 point six and 2.718 to 2 point seven one eight, avoiding Pocket’s internal sentence splitter from treating the decimal period as end-of-sentence.

But a proper fix might involve changing Pocket’s tts_model.py sentence splitting logic.

Let me know if you want a PR or if you have a better idea.

Abhishek8108:
Fixes #162

## What was wrong

The sentence splitter in `split_into_best_sentences` builds its boundary token set by tokenising `.!...?` and then marks every matching period token as a split point. Decimal numbers like `98.6` contain a period that tokenises identically to a sentence-ending period, so the splitter incorrectly breaks them across chunk boundaries:

```
"The average human body temperature is 98.6°F, which is normal."
→ ["98.", "6°F, which is normal."]
```

This produces broken/mangled audio output for any text containing decimals.

## Fix

Add `_normalize_decimals()` which rewrites `digit.digit` patterns to spoken form before tokenisation:

- `98.6` → `98 point 6`
- `3.14` → `3 point 14`
- `2.718` → `2 point 718`

The rewritten form contains no period, so the splitter never sees it as a boundary. Prose punctuation (sentence-ending periods, ellipses, etc.) is not affected — only digit·period·digit patterns match.

This follows the direction suggested by @gabrieldemarmiesse in the issue: use a text normalisation step rather than trying to patch the splitter's boundary detection logic.

## Notes

- No new dependencies — pure `re` from the standard library
- The fix is applied once at the start of `split_into_best_sentences`, before the first tokenisation pass
- Broader number normalisation (e.g. large integers, currency) is left as a follow-on; this PR is intentionally narrow

## Tests

Added to `tests/test_split_sentences.py`:

- `TestNormalizeDecimals` — 6 unit tests covering simple decimals, multiple decimals, prose periods left untouched, integers with trailing periods untouched
- `test_decimal_not_split_into_separate_chunks` — integration test using the tokenizer fixture: the exact example from #162 (`98.6°F`) must not be split
- `test_multiple_decimals_preserved` — multiple decimals in one sentence all normalised

All 8 new tests pass locally alongside the existing suite.

### DIFF
--- a/pocket_tts/models/tts_model.py
+++ b/pocket_tts/models/tts_model.py
@@ -3,6 +3,7 @@ import logging
 import math
 import os
 import queue
+import re
 import statistics
 import threading
 import time
@@ -975,6 +976,22 @@ def _segments_from_boundaries(
     return segments
 
 
+_DECIMAL_RE = re.compile(r"(\d+)\.(\d+)")
+
+
+def _normalize_decimals(text: str) -> str:
+    """Replace decimal numbers with spoken form to avoid spurious sentence splits.
+
+    The sentence splitter treats every period token as a potential boundary.
+    Decimals like '98.6' therefore get incorrectly split into '98.' and '6…',
+    producing broken audio output.  Rewriting to '98 point 6' before tokenisation
+    removes the period from the token stream so the splitter never sees it.
+
+    Only digit·period·digit patterns are rewritten; prose punctuation is untouched.
+    """
+    return _DECIMAL_RE.sub(r"\1 point \2", text)
+
+
 def split_into_best_sentences(
     tokenizer,
     text_to_generate: str,
@@ -985,6 +1002,7 @@ def split_into_best_sentences(
     text_to_generate, _ = prepare_text_prompt(
         text_to_generate, pad_with_spaces_for_short_inputs, remove_semicolons
     )
+    text_to_generate = _normalize_decimals(text_to_generate)
     text_to_generate = text_to_generate.strip()
     tokens = tokenizer(text_to_generate)
     list_of_tokens = tokens.tokens[0].tolist()
--- a/pocket_tts/models/tts_model.py
+++ b/pocket_tts/models/tts_model.py
@@ -977,44 +977,108 @@ def _segments_from_boundaries(
     return segments
 
 
-_DECIMAL_RE = re.compile(r"(\d+)\.(\d+)")
+# Patterns for "atoms" — substrings that contain split-relevant punctuation
+# but should never be treated as sentence boundaries (decimals, URLs, emails,
+# honorifics, common abbreviations).  Order matters in :func:`_atom_patterns_for`:
+# email comes before URL so the URL pattern doesn't consume the domain part of
+# an email; URL comes before the digit pattern so IP-like labels in URLs are
+# captured as one atom.
+_EMAIL_RE = re.compile(r"\b[\w.+-]+@[\w-]+(?:\.[\w-]+)+\b")
+_URL_RE = re.compile(r"\b(?:https?://)?(?:[\w-]+\.)+[a-zA-Z]{2,}(?:/\S*)?")
+# Number with an internal '.' or ',' separator: decimals (98.6, 37,0) and
+# thousands separators (1,000, 1.000) in any locale.
+_DIGIT_NUMBER_RE = re.compile(r"\d+(?:[.,]\d+)+")
 
-# Spoken form of the decimal point for each supported language config stem.
-# Languages not listed here fall back to "point".
-_DECIMAL_WORD: dict[str, str] = {
-    "english": "point",
-    "french": "virgule",
-    "french_24l": "virgule",
-    "german": "Komma",
-    "german_24l": "Komma",
-    "spanish": "coma",
-    "spanish_24l": "coma",
-    "portuguese": "vírgula",
-    "portuguese_24l": "vírgula",
-    "italian": "virgola",
-    "italian_24l": "virgola",
+# Per-language honorifics and common abbreviations.  Keys use the base language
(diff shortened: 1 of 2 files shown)


memchr:
What about languages that [use a comma as a decimal separator](https://en.wikipedia.org/wiki/Decimal_separator#Countries_using_decimal_comma)? Zum Beispiel in Deutsch ist es 37,0 °C

Abhishek8108:
Great point, thanks for catching that! I've updated the PR to address this.

The fix now adds a `_DECIMAL_WORD` mapping covering all language configs that pocket-tts currently ships (English, French, German, Spanish, Portuguese, Italian — including the `_24l` variants). The spoken form is threaded through from `TTSModel.origin.stem` so the correct word is picked automatically based on the loaded model:

- English → `point` (e.g. `98.6` → `98 point 6`)
- German / Dutch-adjacent → `Komma` (e.g. `98.6` → `98 Komma 6`)
- French → `virgule`
- Spanish → `coma`
- Portuguese → `vírgula`
- Italian → `virgola`

Unknown or future languages fall back to `point`. No new dependencies added — just an enumerated dict.

mscdex:
This is at least tangentially related, but perhaps we should at the very least allow supplying a list of sentences instead of just a single string to `generate_audio()`/`generate_audio_stream()`. This way people can perform their own sentence splitting and normalization. If a list of strings is supplied, any/all splitting and decimal or other processing is skipped.

The reason for this is that decimals are not the only cases where there can be issues with the current sentence splitting logic (think of honorifics and such), so allowing end users to delegate handling of that to a separate library would be very beneficial.

On second thought, maybe we alternatively (or additionally) allow people to pass in a list of lists of tokens. Currently `generate_audio()`/`generate_audio_stream()` seemingly unnecessarily calls `prepare_text_stream()` twice on the same input (once to count tokens to create chunks, then again to pass tokens to the model). That would allow users to still (explicitly) apply max_tokens limitations and avoid duplicate work.

Abhishek8108:
Quick follow-up on this — re-reading your comment, I realised the previous "add `_DECIMAL_WORD` per language" change actually only addressed half your point. It picked the right spoken word for each locale, but the regex was still `\d+\.\d+`, so a German user typing `37,0` (the native form you flagged) would have still had it broken on the comma in any sentence long enough to trigger fallback comma-splitting. Apologies for the partial fix.

@mscdex your broader observation is right too — decimals are one symptom of a more general issue: the splitter treats every period/comma/semicolon/colon token as a boundary regardless of context, so honorifics (`Dr. Smith`), abbreviations (`e.g.`), URLs (`example.com`), emails, and thousands separators (`1,000`) all break the same way.

Just pushed a rework that addresses both. Instead of rewriting decimals to spoken form, the splitter now masks every "atom" (decimal, URL, email, honorific, abbreviation) with an opaque letter-only placeholder before tokenisation, and restores the original text in each chunk before returning. Three consequences:

1. **Both decimal forms work in every locale.** `\d+(?:[.,]\d+)+` matches `98.6`, `37,0`, `1,000`, `1.000,50`, `192.168.1.1` — all stay intact regardless of language.
2. **No spoken-word translation needed.** The TTS model already pronounces `"98.6"` and `"37,0"` correctly — only the splitter was confused. Restoring originals lets the model do what it already does well, and the per-language `_DECIMAL_WORD` table goes away.
3. **Adding new atom types is one regex.** Honorifics and abbreviations ship per-language; future "X breaks splitting" reports close in a line or two.

Tests cover all the categories above plus regression guards (English `"Hello, world"` and `"1,000"` still untouched, prose periods still split correctly).

@mscdex on your second suggestion — exposing this as a pluggable `TextNormalizer` interface so callers could swap in domain-specific atoms or skip normalisation entirely — happy to open a separate proposal issue if maintainers would find that useful. Wanted to keep this PR focused on the fix.

gabrieldemarmiesse:
Hello, thanks for the PR! 

Overall, text pre-processing for TTS is a huge subject, with many of corner cases and many languages to support. We take the decision here to let users do the text preprocessing while pocket-tts focus on reading the words in the text. Otherwise I think it could be an entire project to do good text preprocessing.

What we'd love to see is someone making a python package focused on text preprocessing for TTS, but so far, I don't see a good package for that. Maybe it's something that's claude-codable?

So I'll have to close this PR, we want to make this repo as readable as possible, notably for alternative implementations. I understand it's more painful for average users, but on the flip side, we don't do anything too unexpected with the text. So it's easy to preprocess for users. I totally encourage you to start a package with those ideas!

mscdex:
> What we'd love to see is someone making a python package focused on text preprocessing for TTS, but so far, I don't see a good package for that.

From my limited experience, the one from the `nemo-text-processing` Python module seems to do a decent job.

gabrieldemarmiesse:
Yup, nemo is good, it's just painful to install in my experience.

PR #191, opened 2026-05-17
docs: clarify uv virtual environment usage with --active flag

Ronitkapar:
This PR improves the CLI documentation for `pocket-tts` by clarifying the virtual environment mismatch warning that occurs when using `uv`. It explains why the `VIRTUAL_ENV does not match the project environment` warning appears, namely due to `uv` creating its own environment by default, and provides the correct usage with `uv run --active pocket-tts generate`. Additionally, it includes alternative methods (`uvx` and `pip`), clarifies the expected output file (`./tts_output.wav`), and refines wording for better clarity. These documentation updates aim to reduce confusion for users working with custom virtual environments and improve the overall onboarding experience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,21 @@ Navigate to the [Kyutai website](https://kyutai.org/pocket-tts) to try it out di
 ## Trying it with the CLI
 
 ### The `generate` command
-You can use pocket-tts directly from the command line. We recommend using
+You can use pocket-tts directly from the command line.
+
+### ⚠️ Virtual Environment Note
+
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
We'll want to keep the README.md quite lean and to the point. I believe that now, installations issues can quite easily be handled by agents or stackoverflow, which can help the user understand how python virtualenv and uv work. Thank you for the PR though, maybe some users who have the same issue will end up seeing it when they are in trouble

PR #194, opened 2026-05-21
feat(emotts): add EmoTTS prosody parameter steering support

honglac3579:
## Summary
- Added non-invasive, generation-time prosody steering to `EmoTTS`.
- Defined default prosody parameters (`temp`, `noise_clamp`, and `lsd_decode_steps`) for each emotion under `EMOTION_PROSODY_DEFAULTS`.
- Added support for loading custom prosody parameter overrides at instantiation or via `load_model()`.
- Implemented baseline generation parameter capture and restoration to support switching between emotional and neutral states.

## Test Plan
- Comprehensive unit tests in `tests/test_emotts.py` verifying schema validation, parameter snapshotting, state restoration, and override configurations.
- Generation validation in `scripts/test_emotts_cloning.py` verifying that all 5 emotional voice configurations generate distinct audio signatures via checksum validation.

### DIFF
--- a/pocket_tts/emotts.py
+++ b/pocket_tts/emotts.py
@@ -1,0 +1,239 @@
+import json
+import logging
+from pathlib import Path
+from typing import TypedDict
+
+import torch
+import torch.nn as nn
+
+from pocket_tts.default_parameters import (
+    DEFAULT_EOS_THRESHOLD,
+    DEFAULT_LSD_DECODE_STEPS,
+    DEFAULT_NOISE_CLAMP,
+    DEFAULT_TEMPERATURE,
+)
+from pocket_tts.models.tts_model import TTSModel
+from pocket_tts.utils.utils import download_if_necessary
+
+logger = logging.getLogger(__name__)
+
+
+class EmoProsodyParams(TypedDict):
+    temp: float
+    noise_clamp: float | None
+    lsd_decode_steps: int
+
+
+EMOTION_PROSODY_DEFAULTS: dict[str, EmoProsodyParams] = {
+    "angry": {"temp": 1.3, "noise_clamp": 3.0, "lsd_decode_steps": 1},
+    "disgust": {"temp": 0.5, "noise_clamp": 1.5, "lsd_decode_steps": 2},
+    "fear": {"temp": 1.6, "noise_clamp": None, "lsd_decode_steps": 1},
+    "happy": {"temp": 0.9, "noise_clamp": 2.5, "lsd_decode_steps": 2},
+    "neutral": {"temp": 0.7, "noise_clamp": None, "lsd_decode_steps": 1},
+    "sad": {"temp": 0.4, "noise_clamp": 0.8, "lsd_decode_steps": 3},
+}
+
+
+class EmoShiftLayer(nn.Module):
+    EMOTIONS = ["angry", "disgust", "fear", "happy", "neutral", "sad"]
+
+    def __init__(self, hidden_size: int = 1024, injection_layer: int = 5):
+        super().__init__()
+        self.hidden_size = hidden_size
+        self.injection_layer = injection_layer
+        self.steering_vectors = nn.ParameterDict(
+            {emotion: nn.Parameter(torch.zeros(hidden_size)) for emotion in self.EMOTIONS}
+        )
+        self.active_emotion = "neutral"
+        self.intensity = 0.0
+
+    def set_emotion(self, emotion: str, intensity: float):
+        assert emotion in self.EMOTIONS, f"Choose from: {self.EMOTIONS}"
+        assert intensity >= 0.0, "Intensity must be non-negative"
+        self.active_emotion = emotion
+        self.intensity = intensity
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if self.active_emotion == "neutral" or self.intensity == 0.0:
+            return x
+        vec = self.steering_vectors[self.active_emotion]
+        device = x.device
+        dtype = x.dtype
+        vec_shifted = vec.to(device=device, dtype=dtype)
+        return x + self.intensity * vec_shifted.unsqueeze(0).unsqueeze(0)
+
+
+class EmoTTS(TTSModel):
+    """
+    EmoTTS class extending TTSModel to support EmoShift activation steering.
+    Automatically downloads and registers EmoShift weights on top of pocket-tts.
+    """
+
+    def __init__(
+        self,
+        tts_model: TTSModel,
+        weights_path: Path,
+        meta: dict | None = None,
+        prosody_overrides: dict[str, dict] | None = None,
+    ):
+        # Initialize nn.Module first
+        nn.Module.__init__(self)
+
+        # Copy attributes from base model
+        self.__dict__.update(tts_model.__dict__)
+
+        # Instantiate and load EmoShiftLayer
+        device = next(tts_model.parameters()).device
+        self.emo_layer = EmoShiftLayer(hidden_size=1024, injection_layer=5)
+
+        state = torch.load(weights_path, map_location="cpu")
+        self.emo_layer.load_state_dict(state)
+        self.emo_layer.to(device)
+        self.emo_layer.eval()
+
+        self.meta = meta or {}
+        self._hook_handle = None
+        self._register_hook()
+
+        # Capture base generation parameters
+        self._base_params = {
+            "temp": float(self.temp),
+            "noise_clamp": float(self.noise_clamp) if self.noise_clamp is not None else None,
+            "lsd_decode_steps": int(self.lsd_decode_steps),
+        }
+
+        # Initialize prosody profile
+        self._prosody_profile = {
+            emo: params.copy() for emo, params in EMOTION_PROSODY_DEFAULTS.items()
+        }
+
+        # Validate and apply overrides
+        if prosody_overrides is not None:
+            if not isinstance(prosody_overrides, dict):
+                raise TypeError("prosody_overrides must be a dictionary")
+            for emo, params in prosody_overrides.items():
+                if emo not in self.available_emotions():
+                    raise ValueError(
+                        f"Unknown emotion in overrides: '{emo}'. "
+                        f"Choose from {self.available_emotions()}"
+                    )
+                if not isinstance(params, dict):
+                    raise TypeError(f"Overrides for emotion '{emo}' must be a dictionary")
+
+                # Check for required keys
+                for key in ["temp", "noise_clamp", "lsd_decode_steps"]:
+                    if key not in params:
+                        raise ValueError(f"Missing required key '{key}' in overrides for '{emo}'")
+
+                temp = params["temp"]
+                nc = params["noise_clamp"]
+                steps = params["lsd_decode_steps"]
+
+                if not isinstance(temp, (int, float)):
+                    raise TypeError(f"'temp' for '{emo}' must be a float or int, got {type(temp)}")
+                if nc is not None and not isinstance(nc, (int, float)):
+                    raise TypeError(
+                        f"'noise_clamp' for '{emo}' must be float, int, or None, got {type(nc)}"
+                    )
+                if not isinstance(steps, int) or isinstance(steps, bool):
+                    raise TypeError(
+                        f"'lsd_decode_steps' for '{emo}' must be an int, got {type(steps)}"
+                    )
+
+                self._prosody_profile[emo] = {
+                    "temp": float(temp),
+                    "noise_clamp": float(nc) if nc is not None else None,
+                    "lsd_decode_steps": int(steps),
+                }
+
+    def _register_hook(self):
+        if self._hook_handle is not None:
+            self._hook_handle.remove()
+
+        def emo_hook(module, input, output):
+            if isinstance(output, tuple):
+                activations = output[0]
+                steered = self.emo_layer(activations)
+                return (steered,) + output[1:]
+            return self.emo_layer(output)
+
+        self._hook_handle = self.flow_lm.transformer.layers[5].register_forward_hook(emo_hook)
+
+    def set_emotion(self, emotion: str | None, intensity: float = 1.0):
+        """Set the active emotion and scaling intensity, overriding acoustic prosody parameters."""
+        if emotion is None:
+            emotion = "neutral"
+
+        if emotion not in self.available_emotions():
+            raise ValueError(
+                f"Unknown emotion: '{emotion}'. Supported emotions are: {self.available_emotions()}"
+            )
+
+        # Update steering vector layer
+        self.emo_layer.set_emotion(emotion, intensity)
+
+        # Override or restore prosody parameters
+        if emotion == "neutral" or intensity == 0.0:
+            # Restore base parameters
+            self.temp = self._base_params["temp"]
+            self.noise_clamp = self._base_params["noise_clamp"]
+            self.lsd_decode_steps = self._base_params["lsd_decode_steps"]
+        else:
+            # Apply emotion-specific overrides
+            params = self._prosody_profile[emotion]
+            self.temp = params["temp"]
+            self.noise_clamp = params["noise_clamp"]
+            self.lsd_decode_steps = params["lsd_decode_steps"]
+
+    def available_emotions(self) -> list[str]:
+        """Return the list of supported emotions."""
+        return EmoShiftLayer.EMOTIONS
+
+    def cleanup(self):
+        """Remove the PyTorch hook from the model."""
+        if self._hook_handle is not None:
+            self._hook_handle.remove()
+            self._hook_handle = None
+
+    def __del__(self):
+        self.cleanup()
+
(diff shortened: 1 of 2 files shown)


analytically:
+1

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+        nn.Module.__init__(self)
+
+        # Copy attributes from base model
+        self.__dict__.update(tts_model.__dict__)
+
+        # Instantiate and load EmoShiftLayer
+        device = next(tts_model.parameters()).device
+        self.emo_layer = EmoShiftLayer(hidden_size=1024, injection_layer=5)
+
+        state = torch.load(weights_path, map_location="cpu")
```
pixelbits-mk, line 89: EmoShift weights load via pickle-based torch.load from an unofficial, unpinned HF repo
pocket_tts/emotts.py:89–90

EmoTTS downloads its "EmoShift" weights from someone's personal Hugging Face account (Sourajit123/SouraTTS), not the official Kyutai one, and loads them with torch.load. That function can execute arbitrary code hidden in the file. Every other model file in this repo avoids that by using safetensors from a pinned, official source — this is the one exception.

Recommendation: Load with safetensors.torch.load_file (or at least torch.load(..., weights_only=True)) and pin the source to a verified commit.

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+        assert emotion in self.EMOTIONS, f"Choose from: {self.EMOTIONS}"
+        assert intensity >= 0.0, "Intensity must be non-negative"
+        self.active_emotion = emotion
+        self.intensity = intensity
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if self.active_emotion == "neutral" or self.intensity == 0.0:
+            return x
+        vec = self.steering_vectors[self.active_emotion]
+        device = x.device
```
pixelbits-mk, line 60: 🔵 **Informational**: EmoShiftLayer.forward redoes device/dtype conversion and reshaping every call

forward() re-runs vec.to(device, dtype) and two unsqueeze() ops each time the layer-5 hook fires, once per decode step. Over a long generation this repeats hundreds of times, adding avoidable allocations on the hot path.

**Recommendation**: Cache the converted, reshaped vector per active emotion and invalidate it only when set_emotion or the device/dtype changes.

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+        self._base_params = {
+            "temp": float(self.temp),
+            "noise_clamp": float(self.noise_clamp) if self.noise_clamp is not None else None,
+            "lsd_decode_steps": int(self.lsd_decode_steps),
+        }
+
+        # Initialize prosody profile
+        self._prosody_profile = {
+            emo: params.copy() for emo, params in EMOTION_PROSODY_DEFAULTS.items()
+        }
```
pixelbits-mk, line 108: 🟡 **Consider Changing**: Top-level type check on prosody_overrides is never exercised

__init__ raises TypeError when prosody_overrides itself isn't a dict, but the validation test never passes a non-dict value (e.g. a list) to check this path.

**Recommendation**: Add a case in test_emotts_prosody_overrides_validation passing a non-dict prosody_overrides and asserting TypeError.

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+        self.hidden_size = hidden_size
+        self.injection_layer = injection_layer
+        self.steering_vectors = nn.ParameterDict(
+            {emotion: nn.Parameter(torch.zeros(hidden_size)) for emotion in self.EMOTIONS}
+        )
+        self.active_emotion = "neutral"
+        self.intensity = 0.0
+
+    def set_emotion(self, emotion: str, intensity: float):
+        assert emotion in self.EMOTIONS, f"Choose from: {self.EMOTIONS}"
```
pixelbits-mk, line 51: 🔵 **Informational**: Negative-intensity assertion in EmoShiftLayer.set_emotion is untested

set_emotion asserts intensity >= 0.0, but the test only ever passes 0.5, so the assertion path is never actually hit.

**Recommendation**: Add a test calling set_emotion('happy', intensity=-1.0) and assert it raises AssertionError.

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+                        f"'noise_clamp' for '{emo}' must be float, int, or None, got {type(nc)}"
+                    )
+                if not isinstance(steps, int) or isinstance(steps, bool):
+                    raise TypeError(
+                        f"'lsd_decode_steps' for '{emo}' must be an int, got {type(steps)}"
+                    )
+
+                self._prosody_profile[emo] = {
+                    "temp": float(temp),
+                    "noise_clamp": float(nc) if nc is not None else None,
```
pixelbits-mk, line 145: 🔵 **Informational**: Bool-as-int guard for lsd_decode_steps has no test

The code specifically rejects bool values for lsd_decode_steps, beyond the generic type check, but no test passes True/False to confirm this works.

**Recommendation**: Add a validation test with lsd_decode_steps=True confirming it raises TypeError instead of being silently treated as 1.

tests/test_emotts.py
```diff
@@ -0,0 +1,175 @@
...
+    layer.set_emotion("happy", intensity=0.5)
+
+    out_shifted = layer(x)
+    assert not torch.equal(out_shifted, x)
+    # Each value should be increased by 0.5 * 1.0 = 0.5
+    assert torch.allclose(out_shifted, x + 0.5)
+
+
+def test_emotts_loading_and_set_emotion():
+    # Load EmoTTS model (requires HF weights download, which we download cache-only if cached, or live)
```
pixelbits-mk, line 28: 🟡 **Consider Changing**: Override-validation logic is only reachable through network-dependent load_model tests

Almost every assertion about overrides, base-param snapshotting, and set_emotion behavior sits inside tests that skip on HF download failure, so offline/no-cache CI runs get no coverage of this new logic at all, even though the validation itself doesn't need the network.

**Recommendation**: Extract override validation/prosody-profile construction into a pure function, or accept a lightweight fake TTSModel in tests, so this logic can be unit tested without a model download.

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+            self.temp = self._base_params["temp"]
+            self.noise_clamp = self._base_params["noise_clamp"]
+            self.lsd_decode_steps = self._base_params["lsd_decode_steps"]
+        else:
+            # Apply emotion-specific overrides
+            params = self._prosody_profile[emotion]
+            self.temp = params["temp"]
+            self.noise_clamp = params["noise_clamp"]
+            self.lsd_decode_steps = params["lsd_decode_steps"]
+
```
pixelbits-mk, line 187: 🔵 **Informational**: cleanup() idempotency and __del__ path are untested

cleanup() is only called once per test; a second call, or __del__ firing on an already-cleaned model, is never exercised even though __del__ always calls cleanup().

**Recommendation**: Add a test calling model.cleanup() twice and asserting no exception, covering the guarded branch on the second call.

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+    def __init__(
+        self,
+        tts_model: TTSModel,
+        weights_path: Path,
+        meta: dict | None = None,
+        prosody_overrides: dict[str, dict] | None = None,
+    ):
+        # Initialize nn.Module first
+        nn.Module.__init__(self)
```
pixelbits-mk, line 80: 🔴 **Must Fix**: EmoTTS construction aliases the wrapped TTSModel's internal nn.Module dicts

self.__dict__.update(tts_model.__dict__) copies _modules/_parameters by reference, not value. Registering emo_layer or the forward hook then also mutates the original tts_model. Any caller keeping a reference to that model sees it steered too, and EmoTTS.cleanup()/__del__ can pull the hook out from under it.

**Recommendation**: Compose rather than alias: wrap tts_model as a submodule, or deep-copy the relevant dicts, so EmoTTS and the original TTSModel instance don't share mutable state.

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+        )
+        self.active_emotion = "neutral"
+        self.intensity = 0.0
+
+    def set_emotion(self, emotion: str, intensity: float):
+        assert emotion in self.EMOTIONS, f"Choose from: {self.EMOTIONS}"
+        assert intensity >= 0.0, "Intensity must be non-negative"
+        self.active_emotion = emotion
+        self.intensity = intensity
+
```
pixelbits-mk, line 55: 🟡 **Consider Changing**: assert used for input validation gets stripped under python -O

EmoShiftLayer.set_emotion validates via assert, which disappears under -O. Called directly (as the unit test does), an invalid emotion then raises a confusing KeyError from steering_vectors instead of a clear error.

**Recommendation**: Replace the asserts with explicit ValueError checks, matching the style already used in EmoTTS.set_emotion.

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+            self.noise_clamp = self._base_params["noise_clamp"]
+            self.lsd_decode_steps = self._base_params["lsd_decode_steps"]
+        else:
+            # Apply emotion-specific overrides
+            params = self._prosody_profile[emotion]
+            self.temp = params["temp"]
+            self.noise_clamp = params["noise_clamp"]
+            self.lsd_decode_steps = params["lsd_decode_steps"]
+
+    def available_emotions(self) -> list[str]:
```
pixelbits-mk, line 188: 🔵 **Informational**: __del__ can throw AttributeError if __init__ fails before _hook_handle is set

__del__ always calls cleanup(), which reads self._hook_handle, but that attribute is only set partway through __init__ (after torch.load/load_state_dict). If those calls fail, later GC hits an AttributeError, printed as a noisy 'Exception ignored in __del__'.

**Recommendation**: Set self._hook_handle = None before any operation that can fail in __init__, or guard cleanup() with getattr(self, '_hook_handle', None).

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+        """Return the list of supported emotions."""
+        return EmoShiftLayer.EMOTIONS
+
+    def cleanup(self):
+        """Remove the PyTorch hook from the model."""
+        if self._hook_handle is not None:
+            self._hook_handle.remove()
+            self._hook_handle = None
+
+    def __del__(self):
```
pixelbits-mk, line 198: 🔵 **Informational**: Hook cleanup relies on __del__ at GC time

__del__ removing the forward hook isn't guaranteed to run promptly, or safely during interpreter shutdown, so it's not a reliable mechanism for deterministic resource cleanup.

**Recommendation**: Encourage explicit cleanup() calls (e.g. via a context manager) and keep __del__ only as a defensive best-effort fallback.

pocket_tts/emotts.py
```diff
@@ -0,0 +1,239 @@
...
+                        f"'lsd_decode_steps' for '{emo}' must be an int, got {type(steps)}"
+                    )
+
+                self._prosody_profile[emo] = {
+                    "temp": float(temp),
+                    "noise_clamp": float(nc) if nc is not None else None,
+                    "lsd_decode_steps": int(steps),
+                }
+
+    def _register_hook(self):
```
pixelbits-mk, line 149: 🟡 **Consider Changing**: Hook always attaches to layer 5, ignoring EmoShiftLayer.injection_layer

EmoShiftLayer takes an injection_layer argument (set to 5 here), but _register_hook always hooks self.flow_lm.transformer.layers[5], so the parameter is dead and the layer count isn't bounds-checked.

**Recommendation**: Use self.emo_layer.injection_layer to pick the hook target, and bounds-check it against the actual number of transformer layers.

pixelbits-mk (changes requested):
## 🔧 CHANGES REQUESTED

**Found**: 1 Must Fix, 4 Consider Changing, 6 Informational
*Tiers by severity: critical/high → 🔴 Must Fix · medium → 🟡 Consider Changing · low/info → 🔵 Informational*
*Decided by policy rules: `critical-blocks-pr` · `high-security-blocks-pr` · `high-architecture-review` · `judgment-needs-human` · `conflicts-need-human` · `reviewer-advises-feedback` · `manual-publish-requested`*

### What this PR does

This PR adds an EmoTTS wrapper around pocket-tts's TTSModel that steers emotion via an activation-shift hook on transformer layer 5, plus per-emotion generation-parameter overrides (temp, noise_clamp, lsd_decode_steps) with a large accompanying test suite. The declared emotion set checks out end-to-end and the tests are thorough on paper, but two things give pause: the new EmoShift weights are pulled via pickle-based torch.load from an unpinned personal HF repo, breaking the repo's safetensors-only convention, and the way EmoTTS is constructed can silently alias the wrapped model's internal state. Most of the test suite's real assertions also only run when the HF download succeeds.

Reviewed by: 👨‍💻 Developer · ⚡ Performance · 🧪 QA · 🔒 Security · 🏗️ Architecture

*New EmoTTS class uses an unusual pattern (nn.Module.__init__ plus manual __dict__ copy) and hardcodes coupling to an internal transformer layer index via a forward hook. It also uses torch.load on downloaded weights without safety checks, and tests silently skip when network/model download fails, weakening actual coverage.*

### Agent feedback

- 🔒 **Security** — Flags the EmoShift weight download as a pickle-deserialization risk since it comes from an unverified personal HF account.
- 🏗️ **Architecture** — Sees the same weight-loading choice as a break from the codebase's safetensors/pinned-commit convention, plus a risky __dict__-aliasing pattern and a dead injection_layer parameter.
- 👨‍💻 **Developer** — Found the same __dict__-aliasing issue as a functional bug that can steer the caller's original model, plus a couple of minor error-handling nits.
- 🧪 **QA** — Points out that most of the new validation logic is only covered by tests that silently skip offline, alongside a handful of small untested edge cases.
- ⚡ **Performance** — Notes a minor per-forward-call allocation in the steering layer's hot path, rated low impact.

### 🔴 Must Fix

| # | Finding | Recommendation |
|---|---|---|
| 1 | EmoTTS construction aliases the wrapped TTSModel's internal nn.Module dicts<br>`pocket_tts/emotts.py:80` | Compose rather than alias: wrap tts_model as a submodule, or deep-copy the relevant dicts, so EmoTTS and the original TTSModel instance don't share mutable state. |

### Consider Changing & Informational

| # |  | Finding | Recommendation |
|---|---|---|---|
| 1 | 🟡 | Top-level type check on prosody_overrides is never exercised<br>`pocket_tts/emotts.py:108` | Add a case in test_emotts_prosody_overrides_validation passing a non-dict prosody_overrides and asserting TypeError. |
| 2 | 🟡 | Override-validation logic is only reachable through network-dependent load_model tests<br>`tests/test_emotts.py:28` | Extract override validation/prosody-profile construction into a pure function, or accept a lightweight fake TTSModel in tests, so this logic can be unit tested without a model download. |
| 3 | 🟡 | assert used for input validation gets stripped under python -O<br>`pocket_tts/emotts.py:55` | Replace the asserts with explicit ValueError checks, matching the style already used in EmoTTS.set_emotion. |
| 4 | 🟡 | Hook always attaches to layer 5, ignoring EmoShiftLayer.injection_layer<br>`pocket_tts/emotts.py:149` | Use self.emo_layer.injection_layer to pick the hook target, and bounds-check it against the actual number of transformer layers. |
| 5 | 🔵 | EmoShiftLayer.forward redoes device/dtype conversion and reshaping every call<br>`pocket_tts/emotts.py:60` | Cache the converted, reshaped vector per active emotion and invalidate it only when set_emotion or the device/dtype changes. |
| 6 | 🔵 | Negative-intensity assertion in EmoShiftLayer.set_emotion is untested<br>`pocket_tts/emotts.py:51` | Add a test calling set_emotion('happy', intensity=-1.0) and assert it raises AssertionError. |
| 7 | 🔵 | Bool-as-int guard for lsd_decode_steps has no test<br>`pocket_tts/emotts.py:145` | Add a validation test with lsd_decode_steps=True confirming it raises TypeError instead of being silently treated as 1. |
| 8 | 🔵 | cleanup() idempotency and __del__ path are untested<br>`pocket_tts/emotts.py:187` | Add a test calling model.cleanup() twice and asserting no exception, covering the guarded branch on the second call. |
| 9 | 🔵 | __del__ can throw AttributeError if __init__ fails before _hook_handle is set<br>`pocket_tts/emotts.py:188` | Set self._hook_handle = None before any operation that can fail in __init__, or guard cleanup() with getattr(self, '_hook_handle', None). |
| 10 | 🔵 | Hook cleanup relies on __del__ at GC time<br>`pocket_tts/emotts.py:198` | Encourage explicit cleanup() calls (e.g. via a context manager) and keep __del__ only as a defensive best-effort fallback. |

### ✅ What checks out

The emotion set is internally consistent: EMOTION_PROSODY_DEFAULTS defines all six of EmoShiftLayer.EMOTIONS (angry, disgust, fear, happy, neutral, sad), which test_emotion_prosody_defaults_covers_all_emotions explicitly verifies.

PR #213, opened 2026-07-10
docs: correct GPU speedup claim + add CUDA usage notes

moduvoice:
## Motivation

The README and AGENTS.md state that GPU gives no speedup for this model ("We tried running this TTS model on the GPU but did not observe a speedup compared to CPU execution, notably because we use a batch size of 1 and a very small model.").

That holds on hardware with very strong single-thread CPU performance (e.g. Apple Silicon, matching the README's own ~6x-real-time M4 headline). But measured on a cloud x86 VM (4 vCPUs) with a Tesla T4, moving the model to GPU gave a consistent **~2.6x speedup over CPU**:

| Metric | CPU (RTF) | GPU (RTF, `.to("cuda")`) | Speedup |
|---|---|---|---|
| Short text (~2.1-2.3s audio) | 2.306 | 6.286 | ~2.7x |
| Long text (~18s audio) | 2.499 | 6.279 | ~2.5x |

(RTF = real-time factor, higher is faster; min of 4 runs after 1 warmup run, <5% variance, GPU peak VRAM 560.5MB.) So this isn't wrong so much as hardware-dependent, and it's worth documenting for users on weaker/thread-limited CPUs where the GPU is still a meaningful win.

While reproducing this I also hit a few undocumented GPU rough edges worth flagging for anyone who tries `.to("cuda")` themselves:
- `generate_audio()` returns a tensor on the model's device, so the README's own quickstart crashes with `TypeError: can't convert cuda:0 device type tensor to numpy` if you naively add `.to("cuda")` — needs `.detach().cpu().numpy()` instead of `.numpy()`.
- `pip install pocket-tts` / `uv add pocket-tts` (the README's documented install path) can silently fall back to CPU on driver/CUDA version mismatches — you only get a `UserWarning`, not an error, so `torch.cuda.is_available()` quietly returns `False`.
- `serve` and the Docker image have no device option at all and are always CPU-only (only the `generate` CLI has a hidden, undocumented `--device`).
- `quantize=True` (int8) is CPU-only and raises `NotImplementedError` if the model has been moved to CUDA.
- The optional `torchao` quantize extra (`pip install pocket-tts[quantize]`) currently requires `torch>=2.11`, newer than any released stable torch, so installing it breaks `quantize=True` even on CPU.

## Changes

Doc-only, no code changes:
- Added a new "Running on GPU" section to README.md with the measurements above, the `.to("cuda")` + `.detach().cpu().numpy()` snippet, and the caveats listed above.
- Softened/contextualized the "Unsupported features" GPU paragraph to note this is hardware-dependent and point to the new section.
- Updated AGENTS.md's "Device" bullet with the same hardware-dependent nuance and caveats, for coding agents working on this repo.

## Testing

- Docs-only change; ran `uv run pytest -n 3 -v` locally to confirm `tests/test_documentation_examples.py` and the rest of the suite still pass unmodified.
- GPU numbers above were reproduced end-to-end via the repo's public API only (`TTSModel.load_model()` / `.to()` / `.generate_audio()`), no internals touched, on a real Tesla T4 (16GB), driver 550.163.01, torch 2.5.1+cu121.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -105,7 +105,16 @@ This is a pure Python package with Rust extensions in `training/rust_exts/audio_
 
 - **Thread Safety**: The code is NOT thread-safe. Server mode does not support concurrent requests.
 - **Batching**: Batch size is always 1. No batching support currently.
-- **Device**: Defaults to CPU. GPU does not provide speedup for this small model.
+- **Device**: Defaults to CPU. Whether GPU helps is hardware-dependent: on hardware with strong
+  single-thread CPU performance (e.g. Apple Silicon) we did not observe a speedup, but on a cloud
+  x86 VM with a Tesla T4, `.to("cuda")` measured ~2.6x speedup over CPU (RTF ~2.3-2.5x CPU vs.
+  ~6.28x GPU). There is no `device` argument on `TTSModel.load_model()` (the `generate` CLI has an
+  undocumented `--device`; `serve` and Docker have none and are CPU-only). `generate_audio()`
+  returns a tensor on the model's device, so code moved to GPU must call `.detach().cpu().numpy()`
+  instead of `.numpy()`. `quantize=True` is CPU-only and raises `NotImplementedError` on CUDA; the
+  optional `torchao` backend currently requires `torch>=2.11`, newer than any released stable
+  `torch`, so installing it breaks `quantize=True` even on CPU. See README's "Running on GPU"
+  section for details.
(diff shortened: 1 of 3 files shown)


danneauxs:
you can get gpu to produce 44 minutes of audio in 1 min of processing. I am doing that now with my code.

gabrieldemarmiesse (approved):
Thanks for those additions!

danneauxs:
Great stuff with propper batching you can acheive 63.1x or 9 hours of audio i 9.6 min.
Wish I had seen this before I developed my own code might have made it easier or potentially faster.

Issue #203, opened 2026-06-25
[audio.cpp] Optimized GGML based Pocket TTS is 3.22x faster than Python on CUDA


0xShug0:
Hi team,

First, thank you for releasing Pocket TTS. It is an impressive model, and I really appreciate the work that went into making high-quality TTS lightweight and accessible.

I recently implemented Pocket TTS in GGML as part of **audio.cpp**, a C/C++ framework for efficiently running audio models. The framework also includes implementations of several other speech and audio models.

In my benchmarks, the GGML implementation of Pocket TTS runs 3.22× faster than the Python/CUDA implementation.

For RTF:
text len: 6,026-character, 1,028-word
audio len (s): 353.12 
wall time (s): 7.30
RTF: 0.021
x faster than real time: 48.40x

You can try it here:

https://github.com/0xShug0/audio.cpp

I wanted to share the implementation with the Pocket TTS team and community. Feedback, testing, and contributions would be greatly appreciated. I would also be happy to collaborate on improving compatibility, performance, or documentation.

Thanks again for building and releasing such a great model!

jjmlovesgit:
Excited to use this to make Pocket TTS a Browser Extension and local TTS engine with a one click install...

jjmlovesgit:
This version using Pocket TTS below is not the Audio CCP version and has some stall problems I think audio.cpp can help address

https://github.com/user-attachments/assets/846a003d-86c3-4fd1-b28a-c129e39b9ef5

0xShug0:
@jjmlovesgit Thanks for your interest in audio.cpp! Let me know if you run into any issues. Happy to help!

jjmlovesgit:
With the extension the Pocket TTS Voices seamlessly appear in any app the looks for Chrome or Edge browser voices and the Audio.ccp engine provides the TTS voice playback using GPU...  Lets you get great local voices with minimal install fuss.  One extension exposes the voices to all workflows vs separate builds each time.  Audio.ccp is terrific!


<img width="1612" height="1066" alt="Image" src="https://github.com/user-attachments/assets/3835659c-6784-4ba7-9dac-16df5c212cfb" />

https://github.com/user-attachments/assets/41758674-4a1b-4222-b2b9-0dbd74279338

0xShug0:
@jjmlovesgit Great! Would you like to share your project repo link or app link so I can add it to README?

jjmlovesgit:
[...]

PR #214, opened 2026-07-10
feat(scripts): add reproducible inference latency / RTF benchmark

fooSynaptic:
Hi team — thank you for releasing Pocket TTS and making high-quality lightweight TTS so accessible. I'm sharing a small reproducible inference benchmark (RTS on CPU/CUDA)  without changing any model code.

### Summary

This PR adds a lightweight, reproducible inference benchmark for Pocket TTS:

- New script: `scripts/bench_inference.py`
- New tests: `tests/test_bench_inference.py`

The harness uses warm-up + timed runs and reports **RTS** using the same definition as `scripts/evaluate_quantization.py`:

    RTS = audio_duration_sec / wall_clock_sec

Values above 1.0 mean faster than real-time. This does **not** change inference speed—it provides a reproducible Python baseline for CPU/CUDA comparisons, useful for discussions like #203 (GGML vs Python CUDA).

### What changed

- **Added** `scripts/bench_inference.py` — CLI benchmark with optional `--json-out`
- **Added** `tests/test_bench_inference.py` — smoke tests

No changes to model code, CLI defaults, or existing scripts.

### Author benchmark results

Config: `warmup=1`, `timed_runs=3`, voice=`alba`, English model.

| Environment | Device | mean RTS | mean wall | mean audio |
|-------------|--------|----------|-----------|------------|
| Mac | CPU | 6.35× | 0.92s | 5.84s |
| H20 | CPU | 2.95× | 2.00s | 5.89s |
| H20 | CUDA (H20) | 7.79× | 0.74s | 5.76s |

H20 CUDA vs H20 CPU: ~2.65×. 

### How to reproduce

```bash
uv run python scripts/bench_inference.py --warmup 1 --runs 3
uv run python scripts/bench_inference.py --device cuda --warmup 1 --runs 3
uv run pytest tests/test_bench_inference.py -v
```

### Test plan

- [x] `uv run pytest tests/test_bench_inference.py -v`
- [x] Manual CPU/CUDA runs on author machines
- [x] `uv run pytest -n 3 -v` (full suite, optional)
- [x] `uvx pre-commit run --all-files` (per CONTRIBUTING.md)

### DIFF
--- a/scripts/bench_inference.py
+++ b/scripts/bench_inference.py
@@ -28,8 +28,7 @@ from pocket_tts import TTSModel
 
 # Short, fixed text so runs are comparable across machines / PRs.
 DEFAULT_BENCH_TEXT = (
-    "The quick brown fox jumps over the lazy dog. "
-    "Benchmarking inference latency on CPU and GPU."
+    "The quick brown fox jumps over the lazy dog. Benchmarking inference latency on CPU and GPU."
 )
 DEFAULT_VOICE = "alba"
 
--- a/scripts/bench_inference.py
+++ b/scripts/bench_inference.py
@@ -1,0 +1,183 @@
+"""
+Reproducible inference latency / RTF benchmark for Pocket TTS.
+
+Inspired by nanoGPT's bench.py: optional warm-up, then timed runs with sync.
+RTS (real-time speed) matches scripts/evaluate_quantization.py:
+    RTS = audio_duration_sec / wall_clock_sec  (>1 means faster than real-time)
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse:
Thank you for the PR, I don't think we need a benchmark as the  default `uvx pocket-tts generate` already generate numbers showing how fast the generation was. Having a true benchmark is overkill in my opinion. Especially since we let other devs do faster implementations, for example in rust

PR #217, opened 2026-07-11
Fix decimal splitting in sentence chunker

preciousimo:
## Summary
- Skip sentence boundaries when a period separates two digits (e.g. `98.6`, `3.14`)
- Add regression tests for issue #162

## Test plan
- [x] `uv run pytest tests/test_split_sentences.py -v`


manukyutai (approved):

manukyutai:
Thanks for the contribution!

PR #219, opened 2026-07-13
Document CPU-only installation (pip pulls ~3 GB of CUDA wheels by default)

jflaflamme:
The README says pocket-tts is "designed to run efficiently on CPUs" and that it "does not require the gpu version of PyTorch" — but on Linux, `pip install pocket-tts` installs exactly that: the CUDA build of torch plus the whole `nvidia-*` runtime stack.

### Reproduction

```
$ uv pip install --dry-run pocket-tts
Would install 66 packages
 + torch==2.13.0
 + nvidia-cublas==13.1.1.3
 + nvidia-cuda-cupti==13.0.85
 + nvidia-cudnn-cu13==9.20.0.48
 + nvidia-nccl-cu13==2.29.7
 ... (15 nvidia-* packages in total)
```

Summed wheel sizes: **~3.07 GB**, of which torch + the NVIDIA CUDA wheels are ~2.81 GB.

With the PyTorch CPU index, the same install is 47 packages, zero `nvidia-*` wheels, and a 192 MB torch wheel:

```
$ uv pip install --dry-run pocket-tts --extra-index-url https://download.pytorch.org/whl/cpu
Would install 47 packages
 + torch==2.13.0+cpu
```

The repo already routes torch to the CPU index for local development via `[tool.uv.sources]` in `pyproject.toml`, so contributors never see this — only end users installing from PyPI do. This bites people on disk-constrained machines, in slim containers, and in CI.

### This PR

Documentation only, no packaging changes: adds a **CPU-only installation** section to the README covering the `--extra-index-url` flag for pip and the equivalent `[tool.uv.index]` / `[tool.uv.sources]` block for uv, plus a pointer from the CLI install section. Notes that macOS and Windows are unaffected, since their default PyPI wheels are already CPU-only.

### Alternative worth considering

A packaging-level fix (e.g. `cpu` / `gpu` extras) would make the default install small without users needing to know about index URLs. I left it out because moving `torch` out of `dependencies` would break plain `pip install pocket-tts` and `uvx pocket-tts generate` for everyone unless a default is kept, and that trade-off feels like a maintainer call. Happy to follow up with that approach if you prefer it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Navigate to the [Kyutai website](https://kyutai.org/pocket-tts) to try it out di
 You can use pocket-tts directly from the command line. We recommend using
 `uv` as it installs any dependencies on the fly in an isolated environment (uv installation instructions [here](https://docs.astral.sh/uv/getting-started/installation/#standalone-installer)).
 You can also use `pip install pocket-tts` to install it manually.
+On Linux, see [CPU-only installation](#cpu-only-installation) to avoid pulling in the CUDA build of PyTorch.
 
 This will generate a wav file `./tts_output.wav` saying the default text with the default voice, and display some speed statistics.
 ```bash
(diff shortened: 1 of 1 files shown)


jflaflamme:
Thanks — addressed both points in 6664610:

- **Size figures:** no longer stated as bare constants; the text now says the CUDA wheels add "several gigabytes" and qualifies the concrete numbers with the version they were measured on (`with torch 2.13, roughly 3 GB instead of 200 MB`), so they read as an example rather than a guarantee.
- **`uvx` example:** added, since `uvx` is the recommended way to run the CLI:
  ```bash
  uvx --index https://download.pytorch.org/whl/cpu pocket-tts generate
  ```
  Verified end to end — the PyTorch index only carries torch wheels, so everything else falls back to PyPI, and the resolution is 47 packages with `torch==2.13.0+cpu` and zero `nvidia-*` wheels.

gabrieldemarmiesse (approved):
Thanks for the PR!

PR #220, opened 2026-07-14
OpenAI compatible API endpoint (/v1/audio/speech)

bezo97:
Hi, I forked the repo to implement an openai api compatible endpoint so I can try `pocket-tts` with `open-webui`.  
Since many tools work well with this interface, it could be considered to add this to pocket-tts for wider adoption. Feel free to merge my fork if it's useful.

Add openai compatible api endpoints:
- `POST` `/v1/audio/speech`
- served along the existing `/tts` endpoint
- minimal implementation, only `wav` response_format, no streaming
- 'voices-folder' argument: all sound files and custom voice models are read from here and registered as available voices
- including documentation

---

The solution works well with open-webui:
image
If anyone wants to try it, this is the config in the admin settings:
image
(`TTS Voice` field can be any predefined or custom voice name, `TTS Model` field is ignored)

---

Thanks for reviewing, I'm open to resolve change requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,12 @@ pocket-tts serve
 ```
 Navigate to `http://localhost:8000` to try the web interface, it's faster than the command line as the model is kept in memory between requests.
 
+The server also exposes an OpenAI API-compatible endpoint (`/v1/audio/speech`) and supports loading custom voices from a folder:
+```bash
+uvx pocket-tts serve --voices-folder ./my_voices
+```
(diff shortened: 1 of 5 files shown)


gabrieldemarmiesse:
Hi many thanks for the PR, somebody already made a similar PR a while ago, so I'll still use the same answer :)  https://github.com/kyutai-labs/pocket-tts/pull/107

PR #223, opened 2026-07-16
Default the English model's temperature to 0.3

manukyutai:
## Motivation

Human evaluations consistently prefer the English model at temperature **0.3** over the current default **0.7**, and the change is free on every objective axis we measured.

**Dedicated head-to-head studies** (all judgments on this single comparison, crowd-sourced raters, LibriSpeech test-clean prompts):

| study | judgments | temp 0.3 | temp 0.7 |
|---|---|---|---|
| Perceived audio quality | 400 | **214 wins (53.5%)**, Elo 2010 [1986, 2032] | 186, Elo 1985 [1959, 2009] |
| Voice similarity (vs reference) | 300 | **158 wins (52.7%)**, Elo 2009 [1981, 2035] | 142, Elo 1990 [1961, 2018] |

An earlier 4-condition quality study (400 judgments shared across 6 pairs) showed the same direction (+29 Elo for 0.3). The preference is modest but reproduced across independent studies, and no study favors 0.7 on any axis.

**Objective metrics** (full 1127-utterance offline eval, granite ASR, 5 independent runs per cell, mean ± sd):

| | temp 0.3 | temp 0.7 |
|---|---|---|
| WER (whisper-normalized) | **0.82 ± 0.02%** | 0.92 ± 0.04% |
| UTMOS | **4.358 ± 0.004** | 4.318 ± 0.005 |
| speaker similarity (WavLM) | 0.503 ± 0.001 | 0.504 ± 0.001 |
| speech rate | 3.25 w/s | 3.30 w/s |

The WER difference is significant (non-overlapping 95% CIs): 0.3 is slightly *more* intelligible as well.

## Change

- `DEFAULT_TEMPERATURE_FOR_LANGUAGE = {"english": 0.3}` with a resolver mirroring the existing per-language voice/text defaults; all other languages keep 0.7 (untested at 0.3).
- `TTSModel.load_model(temp=None)` and `--temperature` now default to the language's tuned value; explicit values override as before.
- Docs updated.

Note: the `english` substring match also covers `english_2026-01`; the evals were run on the current default English model (`english_2026-04`). Happy to pin it to specific configs if preferred.

https://claude.ai/code/session_01RUch6sYtThkymAUpbTHfuS

### DIFF
--- a/pocket_tts/default_parameters.py
+++ b/pocket_tts/default_parameters.py
@@ -1,9 +1,5 @@
 DEFAULT_LANGUAGE = "english"
 DEFAULT_TEMPERATURE = 0.7
-# Human evals (pairwise, 400 judgments) preferred the English model at a lower
-# temperature: same intelligibility (WER), speaker similarity and speech rate,
-# noticeably better perceived quality.
-DEFAULT_TEMPERATURE_FOR_LANGUAGE = {"english": 0.3}
 DEFAULT_LSD_DECODE_STEPS = 1
 DEFAULT_NOISE_CLAMP = None
 DEFAULT_EOS_THRESHOLD = -4.0
@@ -59,13 +55,6 @@ def get_default_text_for_language(language: str | None) -> str:
         if language is not None and key in language:
             return text
     return DEFAULT_TEXT_FOR_LANGUAGE[DEFAULT_LANGUAGE]
-
-
-def get_default_temperature_for_language(language: str | None) -> float:
-    for key, temperature in DEFAULT_TEMPERATURE_FOR_LANGUAGE.items():
-        if language is not None and key in language:
-            return temperature
-    return DEFAULT_TEMPERATURE
 
 
 def get_default_voice_for_language(language: str | None) -> str:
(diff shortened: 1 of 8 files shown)


pocket_tts/default_parameters.py
```diff
@@ -1,5 +1,9 @@
 DEFAULT_LANGUAGE = "english"
 DEFAULT_TEMPERATURE = 0.7
+# Human evals (pairwise, 400 judgments) preferred the English model at a lower
+# temperature: same intelligibility (WER), speaker similarity and speech rate,
+# noticeably better perceived quality.
+DEFAULT_TEMPERATURE_FOR_LANGUAGE = {"english": 0.3}
```
gabrieldemarmiesse: I think we should add instead in the yaml files (the definitions of the models) the default temperature. Here, we add extra logic, and it's going to get more complicated if each language has its own temperature. The yaml files show should what's specific about each model.

gabrieldemarmiesse (changes requested):

pocket_tts/default_parameters.py
manukyutai: Done — moved to the yamls: `Config.default_temperature` (pydantic, falls back to 0.7) + `default_temperature: 0.3` in `english.yaml`/`english_2026-04.yaml` only, python-side map removed. This also scopes the change away from `english_2026-01`, which we never evaluated at 0.3.

gabrieldemarmiesse (approved):

PR #210, opened 2026-07-09
Fix audio and state edge cases

ShiroKSH:
﻿## Summary

- Fix WAV output headers for seekable files while preserving unseekable streaming behavior.
- Avoid corrupting non-16-bit WAV prompts by using the soundfile path for those files.
- Preserve prepared text prompts during streaming generation and surface generation-thread failures before decoder completion.
- Recognize safetensors state URLs that include revisions or query strings, and keep HTTP cache filenames safe when source URLs include query strings.

## Why

These edge cases could produce WAV files with bogus durations, misread voice prompts, silently lose generation failures, or route valid state URLs through audio decoding.

## Checks

- uv run pytest tests/test_audio.py tests/test_generation_regressions.py tests/test_utils.py tests/test_python_api.py -q
- uv run --with ruff ruff check pocket_tts/data/audio.py pocket_tts/models/tts_model.py pocket_tts/utils/utils.py tests/test_audio.py tests/test_generation_regressions.py tests/test_utils.py
- uv run --with ruff ruff format --check tests/test_audio.py tests/test_generation_regressions.py tests/test_utils.py
- git diff --check

PR #231, opened 2026-07-28
fix: write correct frame count in WAV header

### LINKED ISSUE #232: WAV header reports 1B frames (41666s duration) instead of actual content length
## Bug

`StreamingWAVWriter` unconditionally suppresses header patching (`_patchheader = lambda: None`), so even seekable file outputs keep the 1B-frame placeholder set in `write_header`. Python's `wave` module reads this as **41666 seconds** duration regardless of actual content.

`ffprobe` works (reads data-chunk size), but any tool using stdlib `wave` gets garbage metadata.

## Reproduce

```bash
pocket-tts generate --text "Hello" --output-path test.wav
python3 -c "import wave; w=wave.open('test.wav','rb'); print(f'{w.getnframes()} frames = {w.getnframes()/w.getframerate():.0f}s')"
# Output: 1000000000 frames = 41667s   (actual: ~1.7s)
```

## Root cause

`pocket_tts/data/audio.py`, `StreamingWAVWriter.finalize()`:

```python
self.wave_writer._patchheader = lambda: None  # suppresses ALL header patching
self.wave_writer.close()
```

This was added to prevent crashes on unseekable streams (stdout, HTTP), where `close()` calls `_patchheader()` which calls `seek()` and raises. The guard is legitimate for those paths. But it also suppresses header correction on **seekable** streams (files), where `close()` would naturally write the correct data size via `_datawritten` (tracked internally by every `writeframesraw` call).

## Impact

Tools that use Python's `wave` module (rather than ffprobe) get a duration off by 4 orders of magnitude. This affects audio processing pipelines that read WAV metadata for chunking/trimming, media libraries that display duration, and any downstream code that inspects PocketTTS output via stdlib.

## Recommended fix: gate the suppression on `seekable()` (PR #231)

```python
seekable = hasattr(self.output_stream, "seekable") and self.output_stream.seekable()
if not seekable:
    self.wave_writer._patchheader = lambda: None
self.wave_writer.close()
```

Preserves the crash guard on unseekable streams (`FileLikeToQueue` subclasses `io.IOBase` whose default `seekable()` is `False`; stdout pipes return `False`). Lets seekable streams (files, BytesIO) get the correct header. No manual frame counting: CPython's `_patchheader` uses `_datawritten`.

If there is a concern with touching `_patchheader` directly, an alternative is to write file output through a `BytesIO` intermediary (always seekable, wave patches the header correctly, then dump to disk). Same result, slightly more code, one buffer copy.

paoloantinori:
Closes #232.

## Problem

`StreamingWAVWriter` unconditionally suppressed header patching (`_patchheader = lambda: None`), so even seekable file outputs kept the 1B-frame placeholder. Python's `wave` module reads this as 41666s duration regardless of actual content. `ffprobe` works (reads data-chunk size), but any tool using stdlib `wave` gets garbage metadata.

## Fix

The `wave` module already tracks `_datawritten` internally (via `writeframesraw`); it just needs `close()` to run `_patchheader` on seekable streams. Gate the suppression on `stream.seekable()`:

- **Files** (`BufferedWriter`): `seekable()` returns `True` -> header patched with correct frame count
- **Stdout pipe** (`sys.stdout.buffer`): `seekable()` returns `False` -> suppressed (no crash)
- **HTTP** (`FileLikeToQueue`, subclasses `io.IOBase`): default `seekable()` returns `False` -> suppressed (no crash)

No manual frame counting needed. CPython's `_patchheader` uses `_datawritten`, not `_nframes`.

## Why `seekable()` instead of try/except

`try: tell() except:` works but uses exceptions for flow control. `seekable()` is explicit, documented (`io.IOBase.seekable()`), and all stream types in this codebase implement it correctly by inheritance.

## Verification

```python
# Before:
wave.open('output.wav','rb').getnframes()  # 1_000_000_000 (41666s at 24kHz)

# After (file output):
wave.open('output.wav','rb').getnframes()  # 41280 (1.720s at 24kHz)
```

Matches `ffprobe` exactly (1.720s). Stdout path (`--output-path -`) still works (exit 0).

### DIFF
--- a/pocket_tts/data/audio.py
+++ b/pocket_tts/data/audio.py
@@ -107,8 +107,15 @@ class StreamingWAVWriter:
         self.wave_writer.writeframesraw(bytes(num_silence_samples * 2))
 
         if self.wave_writer:
-            # do not update the header for unseekable streams
-            self.wave_writer._patchheader = lambda: None
+            # On seekable streams (files), let close() patch the header so
+            # the frame count matches the actual data. On unseekable streams
+            # (stdout, HTTP), _patchheader would crash on seek(), so suppress
+            # it there. Players that read data-chunk size get the right
+            # duration either way.
+            try:
+                self.output_stream.tell()
+            except (OSError, AttributeError, ValueError):
+                self.wave_writer._patchheader = lambda: None
             self.wave_writer.close()
 
 


gabrieldemarmiesse:
This is handled by #210 , so there is no need to merge this one. Thank you for opening the PR though, a new release will be done soon so you can use the fix.

Issue #189, opened 2026-05-14
Bug Report: Voice Cloning produces garbled/distorted audio on Windows


DocZet0815:
## Summary
Voice cloning produces severely distorted, unintelligible audio on Windows.
Built-in voices (e.g. `"alba"`) work correctly. The issue is reproducible
on two different Windows machines (x86-64 and ARM/Snapdragon X).

## Environment

- **OS**: Windows 11
- **Hardware tested**:
  - Snapdragon X (ARM)
  - x86-64 (second machine, same result)
- **pocket-tts version**: latest (`pip install pocket-tts`)
- **Python**: 3.12
- **PyTorch**: CPU only

## Steps to Reproduce

```python
from pocket_tts import TTSModel
import scipy.io.wavfile

model = TTSModel.load_model()

# Built-in voice -- works fine
voice_ok = model.get_state_for_audio_prompt("alba")
audio_ok = model.generate_audio(voice_ok, "Hello world.")
scipy.io.wavfile.write("ok.wav", model.sample_rate, audio_ok.numpy())

# Custom WAV voice -- produces distorted output
# WAV file: Mono, 24000 Hz, int16, 146s -- verified correct format
voice_bad = model.get_state_for_audio_prompt("./my_voice.wav")
audio_bad = model.generate_audio(voice_bad, "Hello world.")
scipy.io.wavfile.write("bad.wav", model.sample_rate, audio_bad.numpy())
```

## Observed Behavior

- `ok.wav` (built-in voice): clear, intelligible speech ✅
- `bad.wav` (custom WAV): severely distorted, sounds like static/noise ❌

The custom WAV file has been verified to have the correct format:
- Sample rate: 24000 Hz
- Channels: Mono (1D array)
- dtype: int16
- Duration: 146s
- RMS: 7112 (good signal level)

Tested with:
- Direct WAV path (`./my_voice.wav`)
- Via safetensors export/import (`export_model_state` → `get_state_for_audio_prompt`)

Both produce the same distorted output.

## Expected Behavior

Voice cloning should produce clear speech in the style of the provided audio sample,
similar to the quality of built-in voices.

## Additional Notes

- The issue might by Windows-specific -- would appreciate confirmation
  whether this works on Linux/Mac
- Could this be related to how the voice cloning weights are loaded on Windows?
- Is there a known minimum audio duration or quality requirement for the prompt?

DocZet0815:
[...]

PR #236, opened 2026-08-10
Question: should audio_read handle non-16-bit WAV voice prompts?

tomfluff:
**Heads up: this PR was written with AI assistance (Claude Code). I reviewed the change and ran the tests locally, but please review it with that in mind.**

I hit this with my own recordings and I am genuinely unsure whether it is a bug or a deliberate trade-off, so treat this as a proposal rather than a bug report. Happy to close it, shrink it, or turn it into an issue instead.

## What I observed

`audio_read` decodes every WAV as 16-bit PCM regardless of the file's actual sample width:

```python
samples = np.frombuffer(raw_data, dtype=np.int16).astype(np.float32) / 32768.0
```

Feeding it 24-bit recordings (what my recorder produces by default) gave one of two outcomes:

- `ValueError: buffer size must be a multiple of element size`, when the byte count is odd for int16
- silently decoded noise, when the byte count happens to be even

The first is easy to diagnose. The second is what prompted this PR: the garbage is accepted as a voice prompt and encoded without any error, so `export-voice` writes a `.safetensors` full of noise and `generate` happily produces output from it. I only noticed because one of three same-session takes "succeeded" and sounded wrong. Same path affects uploads to the `serve` web UI, where it surfaces as a 500 with no detail.

## Why this may be intended

The stdlib-only WAV path looks deliberate, keeping `soundfile` an optional dependency for the common case, and 16-bit is what most voice samples are. If that trade-off is intentional then most of this PR is unnecessary, and the smaller fix would be to raise a clear error on unsupported sample widths, or route them to `soundfile`, rather than adding decoders.

I could not find an existing issue for this, so I have no idea whether it affects anyone else or just my recording setup. #189 looked similar (cloning noise, built-in voices fine) but turned out to be a sample-rate problem.

## What this PR does, if the behaviour is unintended

- Respects `wav_file.getsampwidth()` and decodes 8, 16, 24, and 32-bit integer PCM. 8-bit is treated as unsigned centered on 128; 24-bit is widened to int32 by padding the low-order byte, preserving value and sign.
- Falls back to `soundfile` for files the `wave` module cannot parse at all (IEEE float, `WAVE_FORMAT_EXTENSIBLE`) instead of propagating `wave.Error`.
- Leaves the 16-bit path unchanged, so the common case keeps the same stdlib-only behaviour.

Happy to reduce this to just the error-on-unsupported-width version if you would prefer to keep the decoder surface small.

## Tests

Added `tests/test_audio_read.py`: round-trips a sine sweep through all four bit depths, checks stereo downmix at 16 and 24-bit, and covers the float-WAV fallback (skipped when `soundfile` is absent).

Verified against a real 24-bit recording: it now decodes bit-identically to a 16-bit conversion of the same file (peak 0.38521 vs 0.38522).

`uv run pytest -n 3` passes, 41 tests.

### DIFF
--- a/pocket_tts/data/audio.py
+++ b/pocket_tts/data/audio.py
@@ -20,27 +20,55 @@ logger = logging.getLogger(__name__)
 FIRST_CHUNK_LENGTH_SECONDS = float(os.environ.get("FIRST_CHUNK_LENGTH_SECONDS", "0"))
 
 
-def audio_read(filepath: str | Path) -> tuple[torch.Tensor, int]:
-    """Read audio file. WAV uses built-in wave module; other formats require soundfile."""
-    filepath = Path(filepath)
+# Integer PCM sample widths, in bytes, that map directly onto a numpy dtype.
+# 24-bit (3 bytes) has no numpy equivalent and is handled separately.
+_PCM_DTYPES = {1: np.uint8, 2: "<i2", 4: "<i4"}
 
-    if filepath.suffix.lower() == ".wav":
-        # Use built-in wave module for WAV files
-        with wave.open(str(filepath), "rb") as wav_file:
-            sample_rate = wav_file.getframerate()
-            n_channels = wav_file.getnchannels()
-            raw_data = wav_file.readframes(-1)
-            samples = np.frombuffer(raw_data, dtype=np.int16).astype(np.float32) / 32768.0
-            if n_channels > 1:
-                samples = samples.reshape(-1, n_channels).mean(axis=1)
-            return torch.from_numpy(samples).unsqueeze(0), sample_rate
 
-    # For non-WAV formats, use soundfile (optional dependency)
+def _decode_pcm(raw_data: bytes, sample_width: int) -> np.ndarray:
+    """Decode little-endian integer PCM bytes to float32 samples in [-1, 1]."""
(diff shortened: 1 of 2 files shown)


pocket_tts/data/audio.py
```diff
@@ -20,27 +20,55 @@
...
+
+
+def _read_wav_with_stdlib(filepath: Path) -> tuple[torch.Tensor, int]:
+    """Read an integer-PCM WAV file using the standard library only."""
+    with wave.open(str(filepath), "rb") as wav_file:
+        sample_rate = wav_file.getframerate()
+        n_channels = wav_file.getnchannels()
+        sample_width = wav_file.getsampwidth()
+        if sample_width not in (1, 2, 3, 4):
+            raise wave.Error(f"unsupported sample width: {sample_width} bytes")
```
gabrieldemarmiesse, line 56: Thank you for the PR. Indeed the error message currently is not great and we need to do better, but we don't want to have to handle all the corner cases manually. Can you just raise an error on non-16bits wavs and just ask the user to install soundfile to auto-fallback to soundfile? Thanks a lot!

gabrieldemarmiesse (commented):
Thanks for the PR!

gabrieldemarmiesse:
Apologies, this behavior is already handled by #210 , so I'll have to close this PR

PR #254, opened 2026-08-26
Finetune a released model into a new language

manu-lm:
Training a new language from scratch works, but starting from released weights reaches the same word error rate much sooner. Measured on 976h of Czech (ParCzech), four runs differing only in initialisation and learning rate, evaluated every few thousand steps on 1127 sentences with English voice prompts and Whisper as the ASR:

| step | from scratch | ft lr 2e-4 | ft lr 2e-5 |
|---|---|---|---|
| 2k | 326% | **29.5%** | — |
| 4k | 217% | **23.7%** | 101% |
| 10k | 45.7% | **12.0%** | 18.6% |
| 15k | 17.1% | 11.3% | 14.7% |
| 25k | 16.2% | 10.7% | 11.8% |

Everything converges to 10-12%; the finetune just gets there roughly 2.5x sooner. Keeping the scratch learning rate matters: 2e-5 is slower throughout and no better at the plateau, because the text embedding starts random and the backbone has to move with it. (Swings of ~3 points between adjacent checkpoints are normal at this WER, so read the trend rather than single points.)

**`training/modules/builders.py`** — a new language means a new tokenizer, so the text embedding has a different row count than the released weights and `load_state_dict(strict=True)` fails outright. Pretrained tensors whose shape does not match the model are now dropped and initialised fresh, with a log line naming them; in practice that is exactly `conditioner.embed.weight`. Everything else transfers. Same-shape loads keep the strict path.

**`training/configs/lsd_finetune_language.yaml`** — the recipe above, as a config to point at your own tokenizer and manifests.

**`training/README.md`** — one paragraph in the Train section explaining when to reach for it.

### DIFF
--- a/training/README.md
+++ b/training/README.md
@@ -20,21 +20,21 @@ uv run training/scripts/prepare_data.py --hours 200
 Training:
 
 ```bash
-uv run training/train.py training/configs/lsd_scratch.yaml
+uv run training/train.py training/configs/scratch.yaml
 ```
 
 Generate samples:
 
 ```bash
 uv run pocket-tts generate --config pocket_tts/config/english.yaml \
-    --checkpoint runs/lsd_scratch/checkpoint_00050000.pt \
+    --checkpoint runs/scratch/checkpoint_00050000.pt \
     --voice some_speaker.wav --text "Hello there."
 ```
 
 Congratulations, you've trained your first Pocket TTS!
 
 With the 200 hours that we used, you should get intelligible speech, but to get a model that's on par with the production Pocket TTS, you'll want 2000 hours or more.
-You can try to rerun with `--hours 2000` and edit `lsd_scratch.yaml` to point to the new manifest.
+You can try to rerun with `--hours 2000` and edit `scratch.yaml` to point to the new manifest.
 
 Now in detail:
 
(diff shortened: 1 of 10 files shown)


manu-lm:
Reworked per review: `reset_text_embedding` is now an explicit flag rather than shape detection.

On why the sizes differ at all — it is not fundamental, it is a convention mismatch. The released configs set `n_bins: 4000`, `train_tokenizer.py` defaults to `--vocab-size 3999`, and `LUTConditioner` allocates `n_bins + 1` rows for padding. So a tokenizer fitted with our own default gives 4000 rows against the released 4001.

That is also the argument for the flag over detection: fit a 4000-vocab tokenizer and the shapes match exactly, the old embedding loads without complaint, and its rows now index a completely different token inventory. Detection would silently do the wrong thing in precisely the case where it matters most. The flag says what is intended; without it, a shape mismatch now raises an error naming the flag instead of a raw torch traceback.

manu-lm:
`train_tokenizer.py` now defaults to `--vocab-size 4000`, matching the `n_bins` the released configs already carry, so a new tokenizer needs no override at all — the finetune config only points `tokenizer_path` at the new model.

Note this makes `reset_text_embedding` load-bearing rather than incidental: with vocab 4000 the shapes now line up, so without the flag the released embedding would load cleanly onto a different token inventory instead of failing.

training/modules/builders.py
```diff
@@ -166,6 +166,20 @@ def build_models(args: TrainArgs) -> tuple[TrainableTTS, MimiModel, tp.Any]:
         flow_state = {
             k.removeprefix("flow_lm."): v for k, v in state.items() if k.startswith("flow_lm.")
         }
+        dropped: list[str] = []
+        if args.reset_text_embedding:
+            dropped = [k for k in flow_state if k.startswith("conditioner.embed.")]
+            logger.info("starting the text embedding from scratch: %s", dropped)
+            flow_state = {k: v for k, v in flow_state.items() if k not in dropped}
+        else:
```
manukyutai: Drop the else

training/args.py
```diff
@@ -62,6 +62,11 @@ class TrainArgs:
     # If false, only Mimi/tokenizer weights are used and the FlowLM is
     # re-initialized (training from scratch).
     start_from_pretrained: bool = True
+    # Load the pretrained weights but start the text embedding from scratch.
+    # Needed when the tokenizer differs from the one the weights were trained
+    # with -- its rows index a different token inventory, so keeping them would
```
manukyutai: Needed when the tokenizer differs from the one the weights were trained with, e.g. when training for a new language

training/README.md
```diff
@@ -154,6 +154,11 @@ The official Pocket TTS training happens in two steps, corresponding to the two
 - `lsd_scratch.yaml`: trains a 24-layer teacher from scratch
 - `lsd_depth_distill.yaml`: distils that teacher into a 6-layer student. This also bakes in classifier-free guidance (CFG), see [paper](https://arxiv.org/abs/2207.12598) or [explanation](https://youtu.be/iv-5mZ_9CPY?t=1797).
 
+For a new language, `lsd_finetune_language.yaml` starts the teacher from
+released weights instead of from scratch: the text embedding is initialised
+fresh (the tokenizer differs) while the backbone transfers. On Czech it reached
+the same 10-12% WER as a scratch run about 2.5x sooner.
```
manukyutai: the same 10-12% WER -> the same WER

training/configs/finetune_language.yaml
```diff
@@ -0,0 +1,43 @@
+# Finetune a released model into a new language: on 976h of Czech this hit 12%
+# WER by 10k steps where a scratch run was still at 46%, and both settled around
+# 10-12%. Keep lr at 2e-4: the text embedding starts random, so the backbone has
+# to move with it.
+model_config: pocket_tts/config/english_2026-04_24l.yaml
+model_overrides:
+  # The sentencepiece model you fitted on your transcripts. n_bins already
+  # matches train_tokenizer.py's default vocab size, so it needs no override.
+  flow_lm.lookup_table.tokenizer_path: data/tokenizer.model
```
manukyutai, line 9: Create another lsd_finetune.yaml also delete the lsd_ prefix for all configs

manukyutai (approved):

PR #255, opened 2026-08-27
Add Apple Silicon (MPS) training support

michaldobiezynski:
The training README says Mac training is unsupported but bugfix PRs are welcome, so here is single-device training working end-to-end on Apple Silicon via the MPS backend, plus the diagnosis and fix for a memory behaviour that otherwise makes it unusable in practice.

## What's here

- `training/distributed.py`: accept MPS as a single-process training device (torchrun/NCCL paths untouched, still CUDA-only)
- `training/train.py`: enable bf16 autocast and fused AdamW on MPS (both verified working on torch 2.13); release the MPS allocator cache after each optimizer step (see below); log `mps drv`/`cur` memory figures on step log lines when training on MPS
- `training/configs/scratch_mps.yaml`: `scratch.yaml` with the consumer-GPU batching (16 x 4, same effective batch of 64), `compile: false`, and checkpoint/sample cadence suited to multi-day consumer runs
- `training/README.md`: an Apple Silicon section with measured numbers

## The allocator fix

With variable-length batches, the MPS caching allocator keeps a cached buffer set per encountered shape and never returns freed blocks to the OS. On an M5 Pro 64 GB, driver-allocated memory grows to ~72 GB against a flat 6.7 GB of live tensors within five steps; macOS then pages the GPU working set and throughput collapses ~12x. This is the same fragmentation problem `train.py` already handles on CUDA with `expandable_segments:True`, which has no effect on MPS.

Measured on the 200h HiFiTTS-2 quickstart data, 25 steps, batch 16 x 4:

| | driver-allocated | live tensors | sustained it/s |
|---|---|---|---|
| without fix | ~72 GB (paging) | 6.7 GB | 0.008 |
| with per-step `torch.mps.empty_cache()` | 19-24 GB | 6.7 GB | 0.11, no decay |

The per-step release has no measurable cost (steps 1-3 run at the same speed with and without it). Larger micro-batches don't help on this hardware: 32 x 2 sustains 0.065 it/s and 64 x 1 exceeds physical memory on real (uncached) demand.

## Testing

- `pytest training/tests`: 65 passed on macOS (arm64, torch 2.13.0)
- 25-step instrumented benchmarks for the numbers above
- A multi-day 200h from-scratch run is currently in progress on the same machine with this exact branch; loss/grad trajectories over the first thousands of steps match the reference curves in the training README

## Limitations

- Single device only; multi-process training still requires CUDA/NCCL
- `torch.compile` is left off in the MPS config (inductor's MPS path untested here)
- At ~0.11 it/s an M5 Pro is roughly a third of an L4, so this targets development, finetuning and patient from-scratch runs rather than parity with data-centre GPUs

### DIFF
--- a/training/README.md
+++ b/training/README.md
@@ -96,9 +96,14 @@ uv run training/train.py training/configs/scratch_mps.yaml
 `torch.compile` targets inductor's CUDA path and is unproven on MPS. bf16 autocast and
 fused AdamW both work on the torch version the project already requires.
 
-Multi-process training stays CUDA-only, NCCL has no MPS backend. Throughput is well below
-a data-center GPU: run a few hundred steps to measure it/s on your machine before
-committing to a long run.
+Multi-process training stays CUDA-only, NCCL has no MPS backend.
+
+Measured on an M5 Pro (20-core GPU, 64 GB): 0.11 it/s sustained at `batch_size: 16`,
+`grad_accum_steps: 4` — roughly a 23 GB L4 at a third speed, so 200k steps is a
+multi-week run. Useful for development, smoke tests and finetuning experiments; plan
+on an NVIDIA GPU for a full from-scratch run. The step log's `mps drv`/`cur` figures
+flag the paging failure mode: if driver-allocated runs far above current-allocated
+and it/s decays, memory is the problem, not compute.
 
 ## Data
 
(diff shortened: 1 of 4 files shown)


gabrieldemarmiesse:
Thank you for the PR, I believe it's been an oversight in the readme, we don't have the hardware or the time to make sure training works on other platforms than linux, so I'll change the readme to be explicit about it. I hope your code will help other devs who want to train on mac though!

PR #263, opened 2026-08-28
Move training data loading into subprocesses

gabrieldemarmiesse-ai-agent:
1xH100 PCIe, 24 CPU cores, `training/configs/lsd_scratch.yaml` defaults (batch 16 x grad_accum 4), 200h HiFiTTS-2, steps 20-40 of a 40-step run:

| | steps/s | batches/s |
|---|---|---|
| main (`a61b787`, in-process loader) | 0.79 | 3.17 |
| this branch (3 loader subprocesses) | 1.34 | 5.36 |

https://claude.ai/code/session_01AordVctPrGG33UnBZXQUb5

### DIFF
--- a/training/train.py
+++ b/training/train.py
@@ -163,12 +163,12 @@ def main(config_path: str) -> None:
     optimizer, ema, device, rank = run.optimizer, run.ema, run.device, run.rank
     progress, start_step = run.progress, run.start_step
 
-    sp = model.flow_lm.conditioner.tokenizer.sp
-    tokenize = sp.encode
+    sentence_piece = model.flow_lm.conditioner.tokenizer.sp
+    tokenize = sentence_piece.encode
     train_loader = iter(
         SubprocessDataLoader(
             args.data.train_jsonl,
-            sp,
+            sentence_piece,
             args.batch_size,
             mimi.sample_rate,
             mimi.frame_rate,
@@ -176,7 +176,7 @@ def main(config_path: str) -> None:
             args.data.max_voice_prompt_sec,
             rank,
             run.world_size,
-            seed=args.seed + rank,
+            seed=args.seed,
             shuffle=args.data.shuffle,
             num_procs=args.data.loader_procs,
         )
--- a/training/train.py
+++ b/training/train.py
@@ -31,7 +31,7 @@ from torch.nn.parallel import DistributedDataParallel as DDP
 
 from training.args import TrainArgs, dump_args, load_args, save_args
 from training.checkpointing import EMA, latest_checkpoint, load_checkpoint, save_checkpoint
-from training.dataloader import DataLoader, encode_batch
+from training.dataloader import DataLoader, SubprocessDataLoader, encode_batch
 from training.distributed import (
     avg_across_ranks,
     get_rank,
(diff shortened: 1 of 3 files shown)


training/train.py
```diff
@@ -163,11 +163,12 @@ def main(config_path: str) -> None:
     optimizer, ema, device, rank = run.optimizer, run.ema, run.device, run.rank
     progress, start_step = run.progress, run.start_step
 
-    tokenize = model.flow_lm.conditioner.tokenizer.sp.encode
+    sp = model.flow_lm.conditioner.tokenizer.sp
```
manukyutai: sp -> sentence_piece here and in other places

manu-lm (changes requested):
The approach is right — each subprocess posing as rank `rank*P+i` of `world_size*P` keeps the sharding disjoint and complete with very little new code, and the speedup is believable on a CPU-bound host (grad_accum 4 means four loads per optimizer step). Two things need fixing before merge, one decision, two nits.

**1. A dead subprocess silently drops its shard, forever (must fix).** The liveness check only runs inside `except queue.Empty:`. With two surviving feeders the queue is never empty, so if one worker dies you train indefinitely on 2/3 of the corpus with no error and no log line. Check `self._procs` liveness periodically — e.g. every N batches in `__iter__` — not only on starvation. Same class of silent truncation as the `zip(strict=True)` pass in #257.

**2. Tensors through an mp.Queue can exhaust file descriptors (fix or document).** Each `Batch` carries ~5 tensors plus one ragged token tensor per row — at batch 16 that is ~20 shared-memory segments per batch, at batch 128 ~130, through torch's default `file_descriptor` sharing strategy. Long runs on default ulimits hit "too many open files". Either `torch.multiprocessing.set_sharing_strategy("file_system")` in `_feed_queue`, or document the ulimit requirement.

**Decision: default on or off.** `loader_procs: 3` changes the data order of every existing config relative to main (three interleaved shard streams instead of one). Probably the right default given the numbers, but it should be a stated choice — and worth one line in the config comment.

**Nits:** the per-process seeds collide across (rank, proc) pairs — `seed + rank` at the call site plus `seed + i` inside means rank 1/proc 0 and rank 0/proc 1 share a seed; harmless since their shards differ, but `seed + rank * num_procs + i` is what is meant. And `_prefetch` above the new class now looks dead on the train path — worth checking whether it still has a caller.

Also, CI has not run — bot-account PRs need a maintainer's approve-and-run.

training/train.py
gabrieldemarmiesse-ai-agent: Renamed everywhere in ce2eb35: the train.py local, the constructor parameter, and the worker's local.

gabrieldemarmiesse-ai-agent:
All addressed in ce2eb35:

1. **Silent shard loss**: `_check_procs()` now runs every 100 batches as well as on every empty-queue timeout, so a dead worker surfaces within seconds even while the queue stays full.
2. **FD exhaustion**: the workers now call `torch_mp.set_sharing_strategy("file_system")`, so shared-memory segments are named files instead of passed descriptors; batch size no longer multiplies open fds.
3. **Default on**: kept `loader_procs: 3` as the default — the PR description's benchmark is the stated justification (0.79 -> 1.34 steps/s on a 1xH100/24-core box), and yes, this changes the data order of existing configs (three interleaved shard streams). The diff is deliberately comment-free; the rationale lives here and in the commit messages.
4. **Seed collision**: fixed as you suggested — the class derives `seed + rank * num_procs + i` and the call site passes the base seed, so all (rank, proc) pairs are distinct.
5. **`_prefetch`**: still live — `DataLoader.__iter__` uses it, and that path runs both in `validate()` and inside every loader subprocess.

CI needs a maintainer approve-and-run when you get a chance.

manu-lm (approved):
All four points addressed and verified in the diff: the liveness check now runs every 100 batches (a dead worker surfaces in seconds instead of silently shrinking the corpus), `file_system` sharing removes the fd ceiling, seeds are spaced as `seed + rank * num_procs + i` with the call site no longer double-counting rank, and the default has a measured justification in the commit trail.

One non-blocking residual: on `queue.Empty` with all workers alive, `__iter__` loops forever silently — a stalled-but-alive worker (e.g. a hung filesystem) looks like slow training. A warning log line in that branch would make it diagnosable. Fine as a follow-up.

CI still needs a maintainer's approve-and-run before this can merge.

manukyutai (approved):

PR #267, opened 2026-08-28
Refresh the H100 speed table after the dataloader change

manu-lm:
Re-measured training throughput with the subprocess dataloader (#263) on the same recipe as the existing table (effective batch 64, 12k-step runs, steady-state steps/s). The H100 rows improve at every scale: 1.91→2.17, 3.35→3.92, 5.25→6.25, 6.85→8.69 steps/s, with the biggest relative gain at 8 GPUs where the old in-process loader was most CPU-bound. Peak VRAM is essentially unchanged. The L4/L40S rows predate the loader change and are left as-is, with a note.


manukyutai (commented):
Don't change the comment

manukyutai (approved):

PR #269, opened 2026-08-30
Restore --voice option in serve command and support local paths in /tts

### LINKED ISSUE #235: Regression: `serve --voice <model.safetensors>` no longer works with previously exported voice models
I recently came back to Pocket-TTS after not using it for several months. Originally I was running it through uvx, and when I launched it again uvx automatically updated to the latest version.

My previous workflow was:

```bash
uvx pocket-tts serve --voice path/to/voice.safetensors
```

where the `.safetensors` file had been exported with Pocket-TTS in March earlier this year.

This worked well because I could:

1. Start the server once.
2. Open the web interface.
3. Enter text.
4. Generate and immediately play audio using the exported voice model.

That workflow appears to no longer be available. The `generate` command can still use my `.safetensors` files, but that is noticeably less convenient because it writes a .wav file that I then have to open manually, and it doesn't provide the low-latency interactive experience that the server previously did. I can still send a .wav file to the server through the web interface, but that seems slower and less consistent in quality than my safetensors file seemed to be.

From reading the documentation, it seems that the current `serve` command expects voice models to be provided through the YAML configuration, and those configuration entries appear to reference remote URLs rather than local exported `.safetensors` files.

I wasn't able to determine from the documentation whether:

* serving local exported voice models is still supported,
* there is a replacement for the old `--voice` workflow,
* or whether this functionality has intentionally been removed.

If the functionality has moved, could the documentation point users toward the new workflow? Do I need to upload my models to the internet or spin up a local file server in order to let the pocket-tts server load them?

If it has been removed, I'd like to request reconsidering support for local exported voice models in the server. Being able to keep several .safetensors files locally and launch the web interface against one of them made Pocket-TTS extremely fast and pleasant to use. For my use case, that was significantly more convenient than regenerating from a .wav sample or setting up additional infrastructure.

If there is already a supported way to serve local .safetensors models, I'd appreciate an example, as I wasn't able to find one in the current documentation.

m-altaifi:
Fixes #235

### Context & Problem

In PR #180 (commit `7db278e`), default voice resolution per language was introduced. While the `generate` command made `--voice` optional and fell back to `get_default_voice_for_language()`, the `serve` command omitted the `--voice` argument from its Typer options during the refactor. 

Additionally, validation on the `/tts` endpoint restricted `voice_url` strictly to `http://`, `https://`, `hf://`, and predefined voice keys, preventing local file paths or pre-exported `.safetensors` model state files from being used as the server's default voice.

### Summary of Changes

- Restores `--voice` option to the `pocket-tts serve` CLI command.
- If `--voice` is passed to `serve`, uses it as `default_voice`; otherwise falls back to `get_default_voice_for_language(language, config)`.
- Updates `/tts` endpoint validation to accept local audio file paths and `.safetensors` files (`_is_safetensors_source()`, `Path.exists()`).
- Adds unit and endpoint tests in `tests/test_cli_serve.py`.
- Updates `docs/CLI Commands/serve.md` to document the `--voice` option.

### Testing

Ran test suite:
- `uv run pytest tests/test_cli_serve.py -v` (5/5 passed)
- `uv run ruff check .` (passed)

### DIFF
--- a/docs/CLI Commands/serve.md
+++ b/docs/CLI Commands/serve.md
@@ -17,7 +17,7 @@ This starts a server on `http://localhost:8000` with the default voice model.
 - `--host HOST`: Host to bind to (default: "localhost")
 - `--port PORT`: Port to bind to (default: 8000)
 - `--reload`: Enable auto-reload for development
-- `--voice`: Path to audio conditioning file (voice to clone) or exported `.safetensors` voice model. Defaults to a built-in voice chosen from the language.
+- `--voice`: Path to an audio conditioning file (voice to clone). Exported `.safetensors` voice models are not supported by the demo server. Defaults to a built-in voice chosen from the language.
 - `--language`: Language for the TTS model, one of `'english_2026-01'`, `'english_2026-04'`, `'english'`, `'french_24l'`, `'german_24l'`, `'portuguese_24l'`, `'italian_24l'`, `'spanish_24l'` (default: `english`, which is the same model as `'english_2026-04'`). Incompatible with `--config`. The "24l" variants are bigger models, not distilled yet and here only as preview.
 - `--config`: Path to a custom config .yaml — a local path, an `https://` URL, or an `hf://` path. Incompatible with `--language`.
 - `--quantize`: Use int8 quantization for the model (default: False). This can reduce memory usage and increase speed, with minimal impact on audio quality.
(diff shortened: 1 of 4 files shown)


gabrieldemarmiesse:
Thank you for the PR, I have added the argument `--default-voice` to the `pocket-tts serve` command, and that works with safetensors file, this should be enough for most use cases!

PR #270, opened 2026-08-31
Precompute training latents; re-encode only the stitch window

gabrieldemarmiesse-ai-agent:
`training/scripts/precompute_latents.py` encodes each training utterance through the frozen Mimi encoder once. Because the encoder is causal with a finite lookback, the loader can then serve any random prompt/target cut by slicing: it decodes and re-encodes only the measured `stitch_frames` window after the cut (~2.8 s) instead of the full utterance + prompt (~21 s padded). The stitch horizon is measured on real audio at precompute time, not derived from the architecture. **Precomputing the validation manifest is forbidden** — validation must encode audio directly so its metrics stay exact and comparable; the script only processes `train_jsonl` and `train.py` rejects a latents `valid_jsonl`.

Speed, 1xH100 PCIe / 24 cores, `lsd_scratch.yaml` defaults (batch 16 x grad_accum 4), 200h HiFiTTS-2, steps 20-40 of a 40-step run:

| | steps/s | batches/s |
|---|---|---|
| main (`ea740e5`) | 1.59 | 6.34 |
| this branch (precomputed train latents) | 2.94 | 11.75 |

Precompute cost: ~8 minutes per 200h of audio on the same box (decode in a process pool, pipelined with the GPU encode; ~1.2 GB of latents per 200h).

Validation-set numbers (audio-encoded validation, per the rule above): 24-layer teacher trained from scratch for 400k steps at effective batch 64 on 2000h HiFiTTS-2 **with precomputed train latents**, measured on 50 batches of the 2000h validation manifest:

| metric | value |
|---|---|
| loss | 0.0108 |
| flow_loss | 0.0093 |
| eos_loss | 0.0150 |

The same model scores WER 1.08% / speaker sim 0.911 / UTMOS 4.28 on the full LibriSpeech test-clean cross-sentence eval (`--cfg 2.0`), vs the README reference 0.94% / 0.929 / 4.32 for this exact setup.

https://claude.ai/code/session_01AordVctPrGG33UnBZXQUb5

### DIFF
--- a/training/train_utils.py
+++ b/training/train_utils.py
@@ -142,47 +142,60 @@ def write_samples(model, mimi, tokenize, args, run_dir, step, voice_latents, dev
 def ensure_train_latents(args: TrainArgs, mimi, device, rank: int, world_size: int) -> None:
     """Train from precomputed latents, encoding them first if needed.
 
-    Every rank encodes a strided share of the chunks; completion is signaled
-    through the shared filesystem (an NCCL barrier would time out). Latents
-    already on disk are checked against the config's Mimi weights.
+    The latents store is keyed by a hash of Mimi's encode-path weights, so
+    changed weights trigger a fresh precompute instead of silently serving
+    stale latents. Every rank encodes a strided share of the chunks;
+    completion is signaled through the shared filesystem (an NCCL barrier
+    would time out).
     """
-    train_path = Path(args.data.train_jsonl)
-    is_latents = train_path.with_suffix(".meta.json").exists()
-    if args.data.precompute and not is_latents:
-        latents_manifest = train_path.with_name(train_path.stem + "_latents.jsonl")
-        meta = latents_manifest.with_suffix(".meta.json")
-        if not meta.exists():
-            from training.scripts.precompute_latents import precompute_manifest
+    from training.scripts.precompute_latents import mimi_encode_hash, precompute_manifest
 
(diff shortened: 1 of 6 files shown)


training/train_utils.py
```diff
@@ -136,3 +137,52 @@ def write_samples(model, mimi, tokenize, args, run_dir, step, voice_latents, dev
...
+        recorded = json.loads(train_meta.read_text()).get("weights_path")
+        expected = str(load_model_config(args.model_config, args.model_overrides).weights_path)
+        if recorded != expected:
+            raise SystemExit(
+                f"latents in {args.data.train_jsonl} were precomputed with\n"
+                f"    {recorded}\n"
+                f"but the config trains against\n"
+                f"    {expected}\n"
+                "Re-run training.scripts.precompute_latents with this config."
+            )
```
gabrieldemarmiesse: Not a huge fan of this. If another mimi is on the same path (the file is replaced), it means that we get wrong latents silently. I think we need instead a hash of the mimi weights to ensure that if the weights changed, we recompute the latents automatically

gabrieldemarmiesse (changes requested):

gabrieldemarmiesse:
LGTM!

gabrieldemarmiesse (approved):

manukyutai (approved):

PR #272, opened 2026-08-31
README: explain depth+CFG distillation

manu-lm:
Expands the one-line mention of depth_distill.yaml into a short section: what the student learns (teacher backbone activations; with distill_cfg_coef 2.0 the target is the teacher's guided two-pass computation, so cfg is baked in and inference runs a single pass at cfg 1), how the config finds the teacher, and when parity arrives (~50k WER, ~150k sim/UTMOS). A distillation speed table (1 and 8 GPUs, precomputed latents vs audio pipeline) is being measured and lands here before merge.

### DIFF
--- a/training/README.md
+++ b/training/README.md
@@ -152,26 +152,13 @@ uv run torchrun --nproc-per-node 8 training/train.py training/configs/scratch.ya
 The training is configured using a single YAML file.
 The official Pocket TTS training happens in two steps, corresponding to the two YAMLs we provide:
 - `scratch.yaml`: trains a 24-layer teacher from scratch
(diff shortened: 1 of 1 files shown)


training/README.md
manukyutai, line 182: Put the depth distillation table here

training/README.md
```diff
@@ -187,6 +198,14 @@ Precompute stores ~6 GB of latents per 1000 h of audio next to the manifest.
 `data.precompute: false` skips the precomputing, which will start training
 immediately but at slower speeds.
 
+And for the distillation step (`depth_distill.yaml`, same data, reusing the
```
manukyutai: And for the distillation step (`depth_distill.yaml`), here is the speed you can expect:

Table

manukyutai (approved):

PR #275, opened 2026-09-01
Move text conditioning into modules/text_conditioner.py

manu-lm:
Merges `pocket_tts/conditioners/` (a 7-line `base.py` holding `TokenizedText` plus `text.py`) into a single `pocket_tts/modules/text_conditioner.py`, matching the one-file-per-concern layout of the rest of `modules/`. Imports updated across the package, training code, and tests; no behavior change.

`tests/test_split_sentences.py` and `training/tests/test_smoke.py` pass (31 tests).

### DIFF
--- a/pocket_tts/modules/text_conditioner.py
+++ b/pocket_tts/modules/text_conditioner.py
@@ -1,5 +1,4 @@
 import logging
-from typing import NamedTuple
 
 import sentencepiece
 import torch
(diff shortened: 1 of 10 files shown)


pocket_tts/modules/text_conditioner.py
```diff
@@ -1,15 +1,19 @@
...
 import torch
 from torch import nn
 
-from pocket_tts.conditioners.base import TokenizedText
 from pocket_tts.utils.utils import download_if_necessary
 
 logger = logging.getLogger(__name__)
 
 
+class TokenizedText(NamedTuple):
```
manukyutai: Is this class needed?

manu-lm: Removed — conditioners now take plain token tensors.

manukyutai (approved):

PR #276, opened 2026-09-01
Rename DummyQuantizer to Bottleneck

manu-lm:
`DummyQuantizer` projects the encoder output into the latent space and quantizes nothing — `Bottleneck` says what it does. File renamed to `modules/bottleneck.py` accordingly. The `quantizer` attribute on `MimiModel` and the `quantizer` config key are unchanged, so checkpoints and configs load as before.

### DIFF
--- a/pocket_tts/models/mimi.py
+++ b/pocket_tts/models/mimi.py
@@ -3,8 +3,8 @@ import logging
 import torch
 from torch import nn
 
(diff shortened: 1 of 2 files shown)


manu-lm:
Dropping: the quantizer attribute and config key can't follow the rename (checkpoint/config compat), and a half-renamed concept is worse than the old name.

PR #278, opened 2026-09-01
Use tanh GELU approximation in the backbone FFN

manu-lm:
Switches `F.gelu` to `approximate="tanh"` in the transformer FFN.

This is a correctness fix, not just an optimization: the released English weights come from training runs that used the tanh GELU approximation, while this inference code computed exact (erf) GELU — a silent train/serve mismatch. The tanh form also matches what GGML/candle CPU runtimes compute, and is faster on CPU.

Measured cost of the mismatch on the released 6L model (paired forward, real inputs): 0.35% mean relative perturbation of the backbone output (cosine ≥ 0.99998, z-space MSE ~2e-6 — four orders of magnitude below the distillation-calibrated quality-neutral level), which is why it was never audible. Generation regression tests pass unchanged.

### DIFF
--- a/pocket_tts/modules/transformer.py
+++ b/pocket_tts/modules/transformer.py
@@ -39,7 +39,7 @@ class StreamingTransformerLayer(nn.Module):
     def _ff_block(self, x: torch.Tensor) -> torch.Tensor:
         x_orig = x
         x = self.norm2(x)
-        update = self.linear2(F.gelu(self.linear1(x)))
+        update = self.linear2(F.gelu(self.linear1(x), approximate="tanh"))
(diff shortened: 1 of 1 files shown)


manu-lm:
Full paired eval of the released English model (1127 LibriSpeech cross-sentence items, cfg 1.0, temp 0.3, same seed), identical except this one line:

| inference GELU | WER | UTMOS | spk sim |
|---|---|---|---|
| exact (before) | 0.91% | 4.364 | 0.922 |
| tanh (this PR) | 0.84% | 4.363 | 0.923 |

Identical within noise; the tiny WER edge for tanh is consistent with the training-side provenance.

PR #279, opened 2026-09-01
Precompute: keep the mimi encoder compiled when beartype is active

manu-lm:
Latent precompute crashes with `InternalTorchDynamoError: NotImplementedError: UserFunctionVariable()` whenever pocket_tts gets imported (beartype active) before the script's own `POCKET_TTS_NO_BEARTYPE` setdefault runs — e.g. a caller importing `training.modules.builders` first, or train.py's embedded auto-precompute. Dynamo cannot trace beartype's wrappers around the SEANet/conv forwards.

Fix: strip beartype's wrappers (exposed as `__wrapped__`) from the mimi module classes before compiling, so the encoder runs **compiled** in all cases — the compile speedup is the point of this script. `torch._dynamo.config.suppress_errors` is additionally set as a last-resort eager net for anything else dynamo cannot trace.

Verified: a beartype-wrapped module crashes strict compile before unwrapping and compiles strictly (suppress off) after.

### DIFF
--- a/training/scripts/precompute_latents.py
+++ b/training/scripts/precompute_latents.py
@@ -89,6 +89,9 @@ def load_frozen_mimi(config) -> torch.nn.Module:
     mimi.eval()
     for p in mimi.parameters():
         p.requires_grad_(False)
+    # Beartype's wrappers (active unless POCKET_TTS_NO_BEARTYPE=1) are opaque
+    # to dynamo and crash the compile; fall back to eager instead of dying.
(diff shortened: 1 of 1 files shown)


manu-lm:
Withdrawn: both entry points already set POCKET_TTS_NO_BEARTYPE before importing pocket_tts (train.py:13, precompute_latents.py:8), so every supported path compiles. The crash came from a custom wrapper script importing pocket_tts before those setdefaults ran — fixed on the wrapper side by setting the env first.

PR #280, opened 2026-09-01
Implement word-level timestamps

### LINKED ISSUE #66: Word-level timestamps
This issue is to track interest in word-level timestamps (when was each word of the input said?) like we provide in [Kyutai TTS 1.6B](https://kyutai.org/blog/2025-07-03-kyutai-tts-1-6b).

For Pocket TTS, we don't get timestamps natively from the model, but if there is significant interest we could integrate some pipeline to extract these from the text + audio, or direct people to solutions. See also @kevinahm's comment in #37.

dpm63:
Closes #66 

Adds timestamped methods for both streaming and non-streaming generation.
We use the attention patterns of attention heads that align closely with word boundaries.
Example from the 6L English checkpoint (top: token-level attention pattern. bottom: word-aggregated attention patterns, reference and predicted alignment):
image

## Explanation
The heads chosen all have fairly similar behaviour, they attend to the word that's currently being spoken, or, if no word is currently being spoken, they attend to the next word that will be spoken.
We use a silence detector to differentiate between those two cases, everything tested worked about equally well, but I ended up settling on RMS -60db.
We start at the first word, we close a word when the next word has higher aggregated attention than itself. We start the next word on its first non-silent frame.
There's special behaviour for the last word, since it doesn't have a later word to attend to, we close it on finding a silent frame.
If the current word has less than a certain threshold of attention (currently 1e-3), it may trigger a transition if any later word has higher attention than itself. This is meant to deal with cases where a word is skipped because it never had more attention than the current word (mainly happens with very short words). In these cases, we could become stuck on a previous word and be unable to transition further.

I tested the following transition methods as well, but found this one to work best:
- Transition to the word with the highest attention
- Transition to the next word when any future word has higher attention

They both have a similar problem, in long sentences, they are very likely to be thrown off by noise in the attention patterns.

## Accuracy
Accuracy of every head, measured as MAE against crisperwhisper 2.0 small's timestamps:
| Checkpoint          | Head        | Samples | Words | Skip rate | MAE           |
|---------------------|-------------|---------|-------|-----------|---------------|
| English 2026-04     | L3H8        | 552     | 3,841 | 0.0260%   | 44.52 ms      |
| English 2026-04 24L | L14H10      | 204     | 1,874 | 0.0000%   | 46.77 ms      |
| English 2026-01     | L3H8        | 120     | 990   | 0.0000%   | 72.90 ms      |
| French 24L          | L11H9+L17H8 | 81      | 628   | 0.0000%   | 59.62 ms      |
| German              | L3H6        | 57      | 465   | 0.0000%   | 57.48 ms      |
| German 24L          | L3H6+L16H6  | 39      | 354   | 0.2825%   | 67.20 ms      |
| Italian             | L4H0        | 46      | 388   | 0.0000%   | 66.98 ms      |
| Italian 24L         | L4H0+L15H12 | 56      | 491   | 0.0000%   | 117.60 ms     |
| Portuguese          | L3H14       | 57      | 533   | 0.0000%   | 68.56 ms      |
| Portuguese 24L      | L3H14+L15H9 | 62      | 597   | 0.0000%   | 120.62 ms     |
| Spanish             | L2H3        | 55      | 507   | 0.0000%   | 91.58 ms      |
| Spanish 24L         | L6H9        | 61      | 570   | 0.0000%   | 78.25 ms      |

## Limitations
We use the same algorithm for streaming and non-streaming generation, this means we don't make use of future information for predicting timestamps.
Therefore, It'd be possible to substantially improve accuracy of non-streaming timestamps by using something like Dynamic Time Warping.
Timestamped generation adds 0.8ms of overhead to the first frame and 0.2ms to subsequent ones, as measured on my machine (i5-12400F)

### DIFF
--- a/docs/API Reference/python-api.md
+++ b/docs/API Reference/python-api.md
@@ -176,6 +176,48 @@ for chunk in model.generate_audio_stream(voice_state, "Long text content..."):
     # Could save chunks to file or play in real-time
 ```
 
+##### `generate_audio_with_timestamps(model_state, text_to_generate, frames_after_eos=None, copy_state=True)`
+
+Generate complete audio with word timestamps.
+
+**Parameters:** Same as `generate_audio()`
+
+**Returns:**
+- `TimestampedAudio.audio`: Audio 1D tensor with shape [samples]
+- `TimestampedAudio.words`: completed `WordTimestamp` records, in input-word order
+
+**Example:**
+```python
+result = model.generate_audio_with_timestamps(voice_state, "Hello world!")
+for word in result.words:
+    print(word.word, word.start_time, word.end_time)
+```
+
+##### `generate_audio_with_timestamps_stream(model_state, text_to_generate, frames_after_eos=None, copy_state=True)`
+
+Generate audio streaming chunks and word timestamps from text input.
+
+**Parameters:** Same as `generate_audio()`
+
+**Yields:**
+- `AudioChunk`: A chunk of generated audio with its start and end times
+- `WordStart`: The word and timestamp when it starts
+- `WordEnd`: The word and timestamps when it ends
+
+**Example:**
+```python
+from pocket_tts import AudioChunk, WordEnd, WordStart
+
+stream = model.generate_audio_with_timestamps_stream(voice_state, "Hello world!")
+for event in stream:
+    if isinstance(event, AudioChunk):
+        process_audio(event.audio)
+    elif isinstance(event, WordStart):
+        print("start", event.word_index, event.word, event.start_time)
+    elif isinstance(event, WordEnd):
+        print("end", event.word_index, event.word, event.end_time)
+```
+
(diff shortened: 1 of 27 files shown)


gabrieldemarmiesse:
Thanks a lot for the pull request! We believe that this work is very interesting but we're a bit reluctant to bring it in because we don't want to have to maintain it. What we could do is the same thing we did for related projects, meaning they can go in their own repository and we can advertise it in our README.md. Meaning you can add a pull request to add a line to this section: https://github.com/kyutai-labs/pocket-tts#alterative-implementations that says "here is a re-implementation of pocket-tts that can generate word-level timestamps". What do you think?

dpm63:
Understood. I'll close this PR and open a new one updating the readme soon, thanks for taking the time to review it.

PR #284, opened 2026-09-02
Enable ruff's ANN rules and annotate every function

gabrieldemarmiesse-ai-agent:
## Summary

- Add `[tool.ruff.lint] extend-select = ["ANN"]` (flake8-annotations) to `pyproject.toml`, keeping ruff's default rules on, with `suppress-none-returning = true` so a function that has no `return` statement (or only returns `None`) needs no return annotation. The 229 explicit `-> None` across the repo are removed accordingly; ty infers `None` for those on its own.
- Fix the remaining violations: every function in `pocket_tts/`, `training/`, `scripts/`, and the tests now declares its argument types, and its return type when it returns a value. Annotations were written against the code and its callers, and checked with ty (`missing-type-argument = "error"`).
- Conventions: concrete module classes (`FlowLMModel`, `MimiModel`, `TrainableTTS`) over `nn.Module`; `ModelState` for streaming state; `dict[str, Any]` only for JSON-like rows; `Any` never appears bare in a signature except the `__torch_dispatch__` hook and two test helpers that pass `**kwargs` straight through (each carries a `noqa` with a reason).
- Small behaviour-neutral touches needed to satisfy the checkers once signatures were typed: `is_file_like` became a `TypeIs` guard, a couple of `assert x is not None` narrowings, the abstract `FlowType.loss` now declares the concrete return tuple instead of `NoReturn`, `StatefulModule.__init__` no longer `return`s the result of `super().__init__`, and `typing_extensions` is bumped to `>=4.10.0` for `TypeIs` (torch already requires that).

## Test plan

- [x] `uvx ruff@0.12.7 check .` and `ruff format --check` pass
- [x] `uv run ty check` passes
- [x] `pytest ./tests ./training/tests`: 115 passed; `test_generate_with_custom_voice` fails locally on `main` too (gated HF weights, skipped in CI)
- [ ] CI green

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,6 @@ line-length = 100
 extend-select = ["ANN"]
 
 [tool.ruff.lint.flake8-annotations]
-# A function without a `return` (or that only returns None) needs no return annotation.
 suppress-none-returning = true
 
 [tool.ruff.format]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ dev = [
 line-length = 100
 
 [tool.ruff.lint]
-select = ["ANN"]
+extend-select = ["ANN"]
 
 [tool.ruff.format]
 line-ending = "lf"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ dependencies = [
     "sentencepiece>=0.2.1",
     "safetensors>=0.4.0",
     "typer>=0.10.0",
-    "typing_extensions>=4.0.0",
+    "typing_extensions>=4.10.0",
     "fastapi>=0.100",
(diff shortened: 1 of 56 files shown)


pyproject.toml
```diff
@@ -56,6 +56,13 @@ dev = [
 [tool.ruff]
 line-length = 100
 
+[tool.ruff.lint]
+extend-select = ["ANN"]
+
+[tool.ruff.lint.flake8-annotations]
+# A function without a `return` (or that only returns None) needs no return annotation.
+suppress-none-returning = true
```
gabrieldemarmiesse, line 63:
```suggestion
[tool.ruff.lint.flake8-annotations]
suppress-none-returning = true
```

PR #285, opened 2026-09-02
Dataloader: lazy manifest access via a cached offset index

manu-lm:
On large corpora the manifest jsonl reaches tens of GB (per-word alignments), and every spawned loader subprocess held its stride of lines as Python strings — gigabytes of heap per process, multiplied by ranks x loader procs, with long-run heap churn ending in host-RAM OOM kills.

`LazyEntries` replaces the string lists: a uint64 line-offset index is built once (atomic tmp+rename, mtime-invalidated, ~8 bytes/line — 90 MB for an 11M-row manifest) and cached next to the manifest; lines are read on demand through the shared page cache. Interface stays `len()` + `[i] -> str`, so the loader is unchanged; the object pickles cheaply to spawn workers (offsets reload lazily in the child).

Verified: unit tests for stride equivalence, spawn pickling, index reuse and stale-index invalidation; full repo suite passes; a live training run on precomputed latents trains normally with the index in place.

### DIFF
--- a/training/dataloader.py
+++ b/training/dataloader.py
@@ -9,6 +9,7 @@ ranks by line index.
 import json
 import logging
 import multiprocessing.queues
+import os
 import queue
 import random
 import threading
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
@manu-lm I added some stricter type-checking, you should update your pr

manu-lm:
Rebased onto current main; `ty check` passes locally and in CI.

manu-lm:
Now actually rebased onto main (incl. #284's ANN rules) — annotations added on the new class, ty and ruff clean, tests pass. My previous comment jumped the gun mid-rebase; disregard it.

manukyutai (approved):

PR #290, opened 2026-09-02
Add "pocket-tts-timestamped" project to README

dpm63:


gabrieldemarmiesse:
Awesome, I see you've even uploaded a wheel to pypi, that's going to be very easy for new users to try it out! Thanks a bunch for doing this!!!

PR #291, opened 2026-09-03
Add Korean 300M community model to README

seastar105:
## Summary

- Add the community-trained Korean 300M Pocket TTS model to the README
- Include a pinned Hugging Face config and Korean generation example
- Document only the 300M teacher model

close #261

PR #293, opened 2026-09-04
Indonesian

anak10thn:


manukyutai:
Hello, we're very grateful for the work! Would it be possible to make a pull request like this one: https://github.com/kyutai-labs/pocket-tts/pull/291

You can always point to your code in the hugging face description, but merging so much code seems like a hastle to maintain.

Thanks a lot!

anak10thn:
Thanks, that's fair — the maintenance cost of the pipeline code is on you, not
me, and a README entry is the right size for it.

Split it up: #294 is the README entry alone, in #291's shape (+7/-0). The
pipeline stays on my fork and the model card links to it:
https://github.com/anak10thn/pocket-tts/tree/indonesian

Two things in there might be worth having upstream on their own, separately
from anything Indonesian — happy to open PRs, or to leave them if you'd rather
not carry them:

- **fp16 autocast on pre-Ampere GPUs.** `train.py` hardcodes bf16, which Turing
  has no tensor cores for. Measured on the T4 this trained on: a 4096³ matmul
  runs at 19.5 TFLOPS in fp16 against 1.7 in bf16 — an 11x slowdown, and
  `torch.cuda.is_bf16_supported()` returns True there because it counts
  emulation, so nothing warns you. The fix picks the dtype from the compute
  capability and adds a GradScaler that is a passthrough on bf16, so Ampere and
  later are untouched. It needs the non-finite-gradient abort to tolerate
  scaler overflow: 7 skipped steps out of 1191 logged over 87k steps here.
- **`--normalizer basic` for `librispeech.py`.** WER currently goes through
  `EnglishTextNormalizer` unconditionally, which applies English spelling and
  contraction rules to whatever language it is handed.

Closing this one in favour of #294.

PR #294, opened 2026-09-04
Add Indonesian 24L community model to README

anak10thn:
## Summary

- Add the community-trained Indonesian 24-layer Pocket TTS model to the README
- Pinned Hugging Face config with an Indonesian generation example
- Documents only the 24-layer teacher

Replaces #293, which carried the training code as well — following the guidance
there to keep this to a README entry, in the shape of #291. The pipeline lives
on my fork and is linked from the model card rather than merged here:
https://github.com/anak10thn/pocket-tts/tree/indonesian

## The model

A 24-layer teacher finetuned from `english_2026-04_24l` with a fresh Indonesian
tokenizer, on 502 h of LEMAS Indonesian speech. Scored on 153 cross-sentence
pairs over 116 held-out speakers (Whisper-large-v3, language-agnostic
normalizer, `--temp 0.3 --cfg 2.0 --n-steps 1`):

| | value |
|---|---|
| speaker similarity | 0.941 |
| WER | 23.18% |
| UTMOS | 2.63 |
| silent / no-EOS | 0 / 0 |

Two caveats worth stating plainly, both in the model card:

- **WER reads against the corpus, not against your English numbers.** The same
  ASR on the eval set's *real* recordings scores 10.08%, because the reference
  transcripts are subtitle-derived. 23.18% is the gap over that floor.
- **UTMOS is capped by the data, not by training.** LEMAS audio is 16 kHz
  against Mimi's 24 kHz, so there is no energy above 8 kHz to learn. Training
  was stopped at step 87,500 of a planned 250,000 once it plateaued: WER moved
  0.6 points between steps 37,500 and 75,000 and validation loss was flat from
  42,500 on.

Speaker similarity is where it does well — above the released English model
(0.922) on a corpus of thousands of YouTube speakers rather than a few
narrators.

## Testing

The exact command in this diff was run from a cleared Hugging Face cache
against the published wheel via `uvx pocket-tts`, and separately against
v3.1.0 from source. Both generate; the pinned revision resolves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,13 @@ uvx pocket-tts generate \
   --text "안녕하세요. 한국어 음성 합성 모델입니다."
 ```
 
+- [Pocket TTS Indonesian 24L](https://huggingface.co/anak10thn/pocket-tts-indonesian) by [@anak10thn](https://huggingface.co/anak10thn) (community research release):
+```bash
+uvx pocket-tts generate \
+  --config hf://anak10thn/pocket-tts-indonesian/config.yaml@6160f98e7a6e71e3aa3e063e582d7827518c632f \
+  --text "Selamat pagi. Ini model sintesis suara bahasa Indonesia."
+```
+
 Want your model here? Head to the [training Readme](https://github.com/kyutai-labs/pocket-tts/blob/main/training/README.md) to get started!
 
 ## Projects using Pocket TTS
--- a/README.md
+++ b/README.md
@@ -312,10 +312,11 @@ uvx pocket-tts generate \
   --text "안녕하세요. 한국어 음성 합성 모델입니다."
 ```
 
-- [Pocket TTS Indonesian 24L](https://huggingface.co/anak10thn/pocket-tts-indonesian) by [@anak10thn](https://huggingface.co/anak10thn) (community research release):
+- [Pocket TTS Indonesian](https://huggingface.co/anak10thn/pocket-tts-indonesian) by [@anak10thn](https://huggingface.co/anak10thn) (community research release), 6 layers, distilled with guidance baked in:
 ```bash
 uvx pocket-tts generate \
-  --config hf://anak10thn/pocket-tts-indonesian/config.yaml@6160f98e7a6e71e3aa3e063e582d7827518c632f \
+  --config hf://anak10thn/pocket-tts-indonesian/indonesian_6l.yaml@f44e4fc2b2fd79918667a1264e34505ea39f04fa \
+  --eos-threshold -6.0 \
   --text "Selamat pagi. Ini model sintesis suara bahasa Indonesia."
 ```
 
--- a/README.md
+++ b/README.md
@@ -312,7 +312,7 @@ uvx pocket-tts generate \
   --text "안녕하세요. 한국어 음성 합성 모델입니다."
 ```
 
-- [Pocket TTS Indonesian](https://huggingface.co/anak10thn/pocket-tts-indonesian) by [@anak10thn](https://huggingface.co/anak10thn) (community research release), 6 layers, distilled with guidance baked in:
+- [Pocket TTS Indonesian](https://huggingface.co/anak10thn/pocket-tts-indonesian) by [@anak10thn](https://huggingface.co/anak10thn) (community research release), 6 layers:
 ```bash
 uvx pocket-tts generate \
   --config hf://anak10thn/pocket-tts-indonesian/indonesian_6l.yaml@635cde7a28301861b120f57ec4dda8525073017c \
--- a/README.md
+++ b/README.md
@@ -315,8 +315,8 @@ uvx pocket-tts generate \
 - [Pocket TTS Indonesian](https://huggingface.co/anak10thn/pocket-tts-indonesian) by [@anak10thn](https://huggingface.co/anak10thn) (community research release), 6 layers, distilled with guidance baked in:
 ```bash
 uvx pocket-tts generate \
-  --config hf://anak10thn/pocket-tts-indonesian/indonesian_6l.yaml@f44e4fc2b2fd79918667a1264e34505ea39f04fa \
-  --eos-threshold -6.0 \
+  --config hf://anak10thn/pocket-tts-indonesian/indonesian_6l.yaml@635cde7a28301861b120f57ec4dda8525073017c \
+  --eos-threshold -5.0 \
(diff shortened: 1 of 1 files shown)


manukyutai (commented):
Thanks! Have you done the cfg distillation?

anak10thn:
No, I haven't — this is the raw 24-layer teacher, `distill_cfg_coef: 0.0`. It
has the CFG *dropout* from training (`text_dropout` and `voice_dropout` at 0.2)
but guidance was never distilled in.

Your question caught a real problem with what I published. I evaluated at
`--cfg 2.0` through `training/eval/librispeech.py`, but `cfg_coef` only exists
in `training/modules/model.py` — `pocket_tts/` has no guidance anywhere, so
`pocket-tts generate` cannot take that path. Every number on my model card
therefore describes a setting users cannot reproduce; what they actually get is
the unguided one.

Re-measuring at cfg 1.0 now, sweeping `--eos-threshold` again since the
optimum I found at cfg 2.0 need not hold there. I'll correct the model card to
lead with the cfg-1.0 numbers and post them here.

If the gap turns out to be large, the honest move is to run the distillation
before this entry lands — a README pointer to a model whose numbers do not
match what the CLI produces is worse for you than no entry. Happy to hold the
PR until then; say the word if you'd rather I close it and reopen when the
distilled student is up.

manukyutai:
Distilling your favorite guidance should be fast, and you can even try to distill to smaller models, which is the key to make pocket tts super tiny :)

anak10thn:
Correcting myself, and the honest version is less flattering than either thing
I said above.

**My "10.6 points" figure was wrong** — it compared cfg 2.0 to cfg 1.0 at a
fixed `--eos-threshold -2.0`, which is the wrong operating point for cfg 1.0.

**Every sweep I used to pick an eos was unsound.** `--num-items N` takes the
first N items of the list, not a random sample, and I had been sweeping at
`--num-items 60`. Same checkpoint, same setting: eos -4.0 scores 21.66% on
that subset and 34.41% on the full 153.

**And the full-list numbers do not form a curve.** cfg 1.0, step 87,500, 153
pairs:

| eos | corpus WER |
|---|---|
| -3.0 | 23.45% |
| -4.0 | 34.41% |
| -5.0 | 20.82% |
| -6.0 | 37.81% |

14-point swings between adjacent settings. That is not an eos effect. Looking
at the per-item records for the -6.0 run:

- corpus WER 37.81%, but **median per-item WER 17.65%**
- 13 items of 153 are above 50% WER; drop them and the corpus number is 18.99%
- the worst single item scores **1833%** — one generation ran ~18x the
  reference length, and corpus-level WER sums insertions, so a handful of
  runaway generations move the aggregate by more than any setting does

So the aggregate I have been quoting — including the 23.18% on my model card —
is a lottery on how many generations run away, not a measurement of the model.
Three different failure modes are collapsed into it: runaway repetition, early
stopping, and Whisper auto-detecting the wrong language and translating (one
Indonesian reference came back as "thank you", scored 100%, on audio that was
fine).

What I am doing about it, before claiming any number again:

- `--asr-language` so the ASR stops detecting per clip (that one is an eval bug
  charging the model for Whisper's mistake — happy to send it as a small PR if
  you want it upstream)
- report median per-item WER and the count of runaway items alongside the
  corpus figure, since the corpus figure alone is not reproducible
- the eval directory name does not include eos, so runs at different thresholds
  overwrite each other's `records.json` — fixing that before re-measuring

**Suggest you hold or close this PR.** I cannot honestly recommend an
`--eos-threshold` for the example yet, and as written it takes the CLI default
of -4.0, which is the worst of the four I measured. Distillation (24L to 6L,
`distill_cfg_coef: 2.0`, ~2.3 days here) starts shortly; I will re-open with
numbers measured properly, or not re-open if they do not hold up.

Sorry for the noise on your tracker — better here than in your README.

anak10thn:
@manukyutai you were right on both counts — it was fast, and the smaller model
is the one worth publishing. This PR now points at the 6-layer student.

`distill_cfg_coef: 2.0`, 24L to 6L, 50k steps, ~28 h on one T4. Scored on 153
held-out pairs, `--cfg 1.0`, each model at its own best `--eos-threshold`:

| | teacher 24L | **student 6L** |
|---|---|---|
| median WER | 16.67% | **12.50%** |
| speaker sim | 0.927 | **0.938** |
| UTMOS | 2.36 | **2.68** |
| size | 1.27 GB | **438 MB** |
| CPU speed | 0.69x real time | **2.23x** |

The student wins on every robust metric and reaches the teacher's *guided*
quality in the one backbone pass the CLI runs. On CPU the teacher is unusable
and the student is not — thanks for the nudge.

One warning worth passing on: the median cannot see a refusal to speak. Here
eos -7.0 ranks first by median while producing silence for 18 of 153 inputs,
against 2 at -6.0. The example uses -6.0.

Three things in `training/eval/librispeech.py` that cost me a day — happy to
send any as a small PR, or to leave them:

- `--num-items N` takes the first N of the list, not a random sample, so it is
  unsound for sweeping (21.66% on the first 60 items vs 34.41% on all 153).
- `eval_dir_name` omits `--eos-threshold`, so a sweep's runs overwrite each
  other's `records.json` — against what its own docstring promises.
- Whisper pads to exactly 30 s of mel, so a generation reaching `--max-sec`
  yields 3001 frames where it expects 3000 and kills the run.

Command in the diff was run as written from a cleared cache via `uvx`. Next,
following your suggestion: distilling below 6 layers.

README.md
```diff
@@ -312,6 +312,14 @@ uvx pocket-tts generate \
   --text "안녕하세요. 한국어 음성 합성 모델입니다."
 ```
 
+- [Pocket TTS Indonesian](https://huggingface.co/anak10thn/pocket-tts-indonesian) by [@anak10thn](https://huggingface.co/anak10thn) (community research release), 6 layers, distilled with guidance baked in:
```
manukyutai: No need to specify "distilled with guidance baked in", it's the default.

README.md
```diff
@@ -312,6 +312,14 @@ uvx pocket-tts generate \
...
+```bash
+uvx pocket-tts generate \
+  --config hf://anak10thn/pocket-tts-indonesian/indonesian_6l.yaml@635cde7a28301861b120f57ec4dda8525073017c \
+  --eos-threshold -5.0 \
```
manukyutai, line 319: Eos threshold having to be very low can be a consequence of training with trailing silence or other non speech time. We have a trim option in the codebase to help with that, feel free to try it out if you have time.

PR #300, opened 2026-09-08
Decode every queued latent per call in the decoder thread

manu-lm:
The decoder thread used to pop one latent from the queue and run the Mimi decoder on a single 80 ms frame per call. It now takes every latent that is already queued and decodes them in one streaming call, then blocks for the next one.

- Output is unchanged: the streaming decoder is exact for any number of frames (checked sample-identical on seeded generations).
- Latency is unchanged: only frames that are already waiting are batched, so the first frame never waits for a second one.
- `TTSModel.max_decoder_frames_per_call` (default 0 = every queued frame) set to 1 is frame-by-frame decoding.

Effect depends on which side is the bottleneck. With the released model on 4 CPU threads the flow LM is slower than the decoder, the queue rarely holds more than one frame, and speed is unchanged (2.4x real time either way on a Xeon Gold 6430). When the LM outruns the decoder, for instance on a GPU or with a faster LM, the decoder was the ceiling and this lifts it by 30 to 50% in our measurements.

### DIFF
--- a/pocket_tts/models/tts_model.py
+++ b/pocket_tts/models/tts_model.py
@@ -101,6 +101,8 @@ class TTSModel(nn.Module):
     ):
         super().__init__()
         self.flow_lm = flow_lm
+        # 0 = decode every queued frame in one call; 1 = one frame per call
+        self.max_decoder_frames_per_call = 0
         self.temp = temp
         self.sampler_decode_steps = sampler_decode_steps
         self.noise_clamp = noise_clamp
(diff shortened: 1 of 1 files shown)


pocket_tts/models/tts_model.py
```diff
@@ -508,23 +510,48 @@ def _decode_audio_worker(
                 latent = latents_queue.get()
                 if latent is None:
                     break
-                mimi_decoding_input = latent * self.flow_lm.emb_std + self.flow_lm.emb_mean
+                # Decode every frame the generator has queued in one call: the streaming decoder
+                # is exact for any number of frames, and one call over several frames costs far
+                # less on CPU than one call per frame. The first frame never waits (we only drain
+                # what is already there). max_decoder_frames_per_call=1 restores frame-by-frame.
```
manukyutai: "restores" is narrative. max_decoder_frames_per_call=1 is frame-by-frame decoding.

pocket_tts/models/tts_model.py
```diff
@@ -508,23 +510,48 @@ def _decode_audio_worker(
                 latent = latents_queue.get()
                 if latent is None:
                     break
-                mimi_decoding_input = latent * self.flow_lm.emb_std + self.flow_lm.emb_mean
+                # Decode every frame the generator has queued in one call: the streaming decoder
```
manukyutai: # Decode every latent frame the generator has queued in one call. The first frame never waits.

PR #309, opened 2026-09-14
Length-bucketed batches by default

manu-lm:
Random batches of ~10 s utterances (HiFiTTS-2: mean 10 s, max 20 s) pad about 40% of the computed frames. The loader now pools 20 batches of samples, sorts them by target length and batches neighbours, then shuffles the batch order; padding drops to ~1%.

Measured on one H100, 24-layer teacher recipe, batch 64, 400 steps: 5.05 -> 6.2 steps/s, loss curves identical at every logged step. Peak memory rises (some batches hold only long utterances): 51 GiB instead of 41.5 GiB for the 1-GPU teacher run.

- `data.bucket_batches` (default 20; 0 restores plain shuffled batches), threaded through `DataLoader` and `SubprocessDataLoader`
- collation refactored into `DataLoader._collate` (raw and precomputed-latent paths unchanged)
- the trainer logs peak GPU memory at the end of a run
- README: bucketing paragraph; the timing tables still show the pre-bucketing numbers and will be refreshed from a same-node measurement of all rows (running now)


training/README.md
```diff
@@ -166,9 +166,20 @@ If you train on 2k hours of HiFiTTS-2 and batch size 64, this is the metrics pro
 
 See [Evaluate](#Evaluate) for more info about the metrics.
 
+Batches are length-bucketed by default (`data.bucket_batches: 20`): the
```
manukyutai: Remove this paragraph

training/args.py
```diff
@@ -28,6 +28,11 @@ class DataArgs:
     # network storage (extra IO threads do not help), and a rank consumes
     # batch_size x steps/s: 6 keeps a small model at 35 it/s x 16 fed.
     loader_procs: int = 6
+    # Length bucketing: pool this many batches of samples, sort them by target
+    # length and batch neighbours together, so rows in a batch have similar
+    # lengths and little padding (random batches of 10 s utterances waste ~40%
+    # of the computed frames). 0 keeps plain shuffled batches.
+    bucket_batches: int = 20
```
manukyutai: rename to num_bucket_batches

training/dataloader/loader.py
```diff
@@ -314,31 +316,59 @@ def _batches(self) -> Iterator[Batch]:
                 chunk_entries = [self.get_entry(i) for i in chunk]
                 got = [s for s in pool.map(self._sample_or_none, chunk_entries) if s is not None]
                 samples.extend(got)
+                if self.num_bucket_batches:
```
manukyutai: Would it make sense to change this to if self.num_bucket_batches > 1 and make 1 the code default instead of 0

training/train.py
```diff
@@ -306,6 +307,8 @@ def _request_stop(signum: int, frame: FrameType | None):
             args.run_dir, args.max_steps, model, optimizer, ema, args.num_ckpt_keep, mimi
         )
         progress.log("checkpoint", args.max_steps)
+        if device.type == "cuda":
```
manukyutai, line 310: Is this change intended?

manu-lm: Yes: it is how the peak VRAM/GPU column of the README timing tables gets measured (bucketed batches raise the peak, since some batches hold only long utterances). One log line at the end of the run, rank 0 only. Happy to drop it if you would rather keep train.py untouched.

manu-lm:
Timing tables replaced: the previous version of this PR carried numbers measured on my local tree with a non-public recipe. The tables now come from this branch with the public scratch.yaml / depth_distill.yaml recipes on the same 2 kh subset the README used before, one exclusive 8xH100 node per configuration, 400 steps, plain vs bucketed. Multi-GPU speeds are lower than the 31 Aug numbers even with bucketing off (8 GPUs: 9.1 vs 10.6); on the 31 Aug commit the same node gives 9.8 today, so part is the environment and part is spread over the September commits; none of it is this change.

manukyutai (approved):

PR #314, opened 2026-09-18
Training: Muon as an optional optimizer

manu-lm:
Adds Muon as an optional optimizer for training: `optim.type: muon` puts the backbone's 2D weights under Newton-Schulz orthogonalized momentum (the fused q/k/v projection is orthogonalized as three blocks) and keeps AdamW for embeddings, gains and the sampler head. `muon_lr` (default 0.005) is the Muon step size, the schedule applies to both parts, and weight decay is rescaled so the per-step decay matches AdamW's. `muon_head: true` extends Muon to the sampler head's 2D weights; `muon_rms_match: true` scales updates by 0.2 * sqrt(max(rows, cols)) (Liu et al. 2025) so one learning rate serves both parts. Both off by default. AdamW stays the default.

Implementation notes: one optimizer object carries both update rules (checkpointing unchanged). The polar factor comes from Polar Express (optimal degree-5 polynomials, 6 steps, bf16, compiled): on 24 matrices of 1024x4096 it lands singular values in 0.992-1.008 (mean |sv-1| 0.003) in 7.1 ms, where 5 Newton-Schulz iterations gave 0.68-1.15 (0.135) in 8.3 ms. Fused weights are orthogonalized per logical block (q, k, v of the attention projection with GQA-sized k/v; adaLN chunks under `muon_head`); orthogonalization is batched per block shape and, under DDP, sharded across ranks with one all_gather per shape. The AdamW side uses torch's fused kernel. Optimizer step on the 24-layer teacher, 4 GPUs: 15.6 ms vs 3.5 ms for fused AdamW. Cost on the 24-layer teacher recipe: 5.3 / 7.3 / 9.1 / 9.8 steps/s at 1 / 2 / 4 / 8 GPUs vs 6.3 / 8.1 / 9.8 / 10.4 for AdamW (6-17%). Unit tests cover the orthogonalization and the two update paths.

What it buys, measured on HiFiTTS-2:
- Full corpus, 24-layer teacher, 400k steps: same step to reach UTMOS 4.0 (175k) and the same endpoint (4.280 vs 4.272 over 3 seeds), so no gain for the default recipe.
- 2000 h subset, 64 rows: UTMOS 4.0 reached 50-75k steps earlier than AdamW (150k vs 200-225k), 4.27 vs 4.05-4.16 at the end of the same 250k cosine.
- Depth distillation of the 6-layer student: the plateau is reached in 5-10k steps instead of ~20k; same endpoint.


training/README.md
```diff
@@ -189,6 +189,10 @@ can expect:
 | 1 x H100 | 64 | 10.5 | 7.5 | 9.8 GiB | ~5.3 h |
 | 8 x H100 | 8 | 21.3 | 21.4 | 3.8 GiB | ~2.6 h |
 
+### Optimizer
```
manukyutai: Remove added documentation

training/train.py
```diff
@@ -167,6 +160,65 @@ def setup(config_path: str) -> Run:
     )
 
 
+def build_optimizer(
```
manukyutai: Can this be moved in train utils or other a relevant file?

training/train_utils.py
```diff
@@ -229,3 +229,62 @@ def is_fresh() -> bool:
...
+        return torch.optim.AdamW(
+            model.parameters(),
+            lr=args.optim.lr,
+            betas=args.optim.betas,
+            eps=args.optim.eps,
+            weight_decay=args.optim.weight_decay,
+            fused=device.type == "cuda",
+        )
+    assert args.optim.type == "muon", args.optim.type
+    from training.modules.muon import MuonWithAuxAdam
```
manukyutai: import at the top

training/train.py
manu-lm: Moved to training/train_utils.py (3ab4f54); train.py only imports build_optimizer. It sat in train.py because I copied the layout from the experimental tree, which has no train_utils module.

PR #317, opened 2026-09-21
Load tokenizers from tokenizer.json, keeping .model support

manu-lm:
Every released tokenizer is now also published as `tokenizer.json` (HuggingFace `tokenizers`), sitting next to the SentencePiece `.model` it was converted from, and the shipped configs point at the JSON.

### Backward compatibility

The `.model` files stay on the hub, untouched. `build_tokenizer()` picks the backend from the config's `tokenizer` field (declared since the first release but never actually read) and falls back to the path suffix, so a community model whose config names a `.model` keeps loading through sentencepiece with no change on their side.

Checked against two real community fine-tunes: `mehdi-hf/pocket-tts-farsi-v2` (note the non-standard `tokenizer_ph.model` filename) generates fine on this branch, and `kaanhgunay/pocket-tts-tr` fails identically here and on `main` for an unrelated reason (`download_if_necessary` rejects tag-style revisions like `@v0.1-base`).

### The conversion is exact

Both hub repos got 16 JSON files each, converted from the `.model` at the same path. Every one round-trips to identical token ids, including the `add_dummy_prefix` cases (leading whitespace, whitespace-only, repeated spaces) where SentencePiece and `tokenizers` can legitimately disagree: the converter restores the unconditional prepend with a `Prepend` normalizer ahead of `Metaspace`. `tests/test_tokenizer_backends.py` asserts id-level and decode-level parity, that each backend is selected as expected, and that both survive the trip to a loader subprocess.

### `train_tokenizer.py` now defaults to unigram

It defaulted to `bpe` while every released tokenizer is `unigram`, so the recipe for adding a language produced a tokenizer from a different family than the ones we ship, and a BPE `.model` cannot be represented as a `tokenizers.Unigram` at all, which breaks the conversion above. The flag still accepts `bpe` and `char`.

### Notes

- `tokenizers` was already in `uv.lock` as a transitive dependency, so the lock diff is two lines promoting it to a direct one.
- Call sites that reached through `tokenizer.sp` now use `tokenizer.encode` / `tokenizer.decode`, which both backends expose; the dataloader workers rebuild the tokenizer through `serialize()` / `encoder_from_serialized()` rather than the sentencepiece-only `serialized_model_proto()`.
- All 81 tests pass, plus `ty` and pre-commit.


pyproject.toml
```diff
@@ -10,6 +10,7 @@ dependencies = [
     "torch>=2.5.0",
     "pydantic>=2",
     "sentencepiece>=0.2.1",
+    "tokenizers>=0.21",
```
gabrieldemarmiesse, line 13: @manu-lm is the sentencepiece tokenizer still used? Do we still need the sentencepiece dependency?

manu-lm: Yes, still used, in two places.

At inference, any config whose `tokenizer_path` names a `.model` still loads through sentencepiece, and that is the whole backward-compatibility story of this PR: community fine-tunes point at a `.model` in their own HF repo, which we cannot update for them. I checked two of them on this branch, `mehdi-hf/pocket-tts-farsi-v2` and `kaanhgunay/pocket-tts-tr`, and both are in that situation.

At training time, `train_tokenizer.py` trains the `.model` in the first place and `convert_tokenizer.py` reads it and checks its own output against it, so both need the package regardless.

I did try making it an optional extra with a lazy import and a "pip install 'pocket-tts[sentencepiece]'" error, but backed it out: a fresh `pip install pocket-tts` would then fail on exactly those community models, which felt like a worse trade than 3 MB (sentencepiece is 3.0 MB installed, against 12 MB for tokenizers, which is now required).

Happy to make it an extra if you would rather have the smaller default install; it's a small diff and I have it ready.

gabrieldemarmiesse: sounds good then, good bot

gabrieldemarmiesse (approved):

PR #318, opened 2026-09-22
Refuse manifests without word alignments at startup

manu-lm:
The training loader cuts each utterance at a word boundary: audio before the cut becomes the voice prompt, audio after it the target, paired with the remaining words. That needs the `words` field that `training/scripts/align_data.py` adds.

Without it the loader used to fall back to a window of the target utterance as the prompt, and with `max_voice_prompt_sec: 0` (no cap) to the *whole* target. A model trained on that learns to copy its prompt and never couples to the text. The failure is invisible in the loss and only shows up at eval: fluent speech, right voice, **98% WER**. It cost a day here, on a manifest that had every other field right.

The fallback is removed, at the one place alignments are needed: `_choose_cut` raises `UnalignedEntry` for an entry without `words`. Two things had to change for that raise to actually stop a run rather than get swallowed:

- the per-sample handler that skips unreadable audio lets `UnalignedEntry` through;
- `_prefetch` forwards exceptions from the loader thread to the consumer, where before an exception in the thread left the consumer blocked on the queue forever.

So the error surfaces with a traceback: in-process for the validation loader, and through the existing dead-worker check for the training subprocesses. The test that asserted the fallback now asserts the refusal.


training/train.py
manukyutai, line 1: Do we need this change?

training/dataloader/loader.py
```diff
@@ -72,6 +72,16 @@ def __init__(
         self.jsonl = jsonl
         self.num_bucket_batches = num_bucket_batches
         self.entries = load_entries(jsonl, rank, world_size)
+        # Cutting an utterance into prompt and target needs word alignments; without
+        # them the prompt would be a window of the target and the model learns to
+        # copy it instead of reading the text. Fail here rather than train that.
+        for i in range(min(len(self.entries), 64)):
```
manukyutai: Why checking this here? can't it be checked where "words" is needed?

training/train.py
```diff
@@ -92,17 +92,6 @@ def setup(config_path: str) -> Run:
         logger.info(f"torch {torch.__version__} | {device} ({gpu}) | world size {world_size}")
         logger.info(f"resolved config from {config_path}:\n{dump_args(args).rstrip()}")
 
-    for name, path in (
```
manukyutai, line 95: Why removing this?

manukyutai (approved):

PR #319, opened 2026-09-23
Ignore EOS on the first 6 generated frames

manu-lm:
Short texts sometimes end before the word is spoken. Before speech starts, the EOS logit of some voices already crosses the threshold (it sits around -3 during a leading pause, and the default threshold is -4), so generation stops during the pause.

This change ignores EOS on the first 6 generated frames (0.48 s). A spoken word never ends that early, and sentences never trigger EOS that soon, so longer outputs are unchanged.

Measured through the public API, 40 single words x 2 seeds x 8 voices, whisper-large-v3, exact-word match:

| model | before | after | sentence WER, 20 sentences (before = after) |
|---|---|---|---|
| english (english_2026-09) | 90.8% | 92.3% | 3.97% |
| english_2026-04 | 90.9% | 92.3% | 4.27% |

Waiting 10 frames hurts English single words (85.5% on english_2026-09), because the extra frames get filled with a second word; 6 helps on both models.


pocket_tts/models/tts_model.py
```diff
@@ -84,6 +84,9 @@ def stamp_state_names(tts_model: "TTSModel"):
 class TTSModel(nn.Module):
     _TOKENS_PER_SECOND_ESTIMATE = 3.0
     _GEN_SECONDS_PADDING = 2.0
+    # EOS is ignored on the first frames: before speech starts, the EOS logit of some voices and
```
manukyutai: Don't mention the french 24l, it's going to be updated today anyways

manukyutai (approved):

PR #320, opened 2026-09-23
Keep the data loaders alive through SIGTERM so the trainer can checkpoint

manu-lm:
The checkpoint-on-SIGTERM path from #277 does not survive a real `scancel`. Slurm signals every process of the job, and the loader subprocesses (plus the `torch_shm_manager` they spawn) died on SIGTERM immediately. The trainer's next batch fetch then failed with `RuntimeError: Connection reset by peer` in `rebuild_storage_filename`, before it reached the stop check, so no checkpoint was written and the run resumed from its last periodic checkpoint. We lost 4.5k steps this way on a 4-GPU run.

The loader processes now ignore SIGTERM and SIGUSR1 (the manager inherits that), so the trainer finishes its step and checkpoints as intended. Since multiprocessing stops daemon children with SIGTERM at exit, the trainer now stops them itself with `SubprocessDataLoader.close()` on both the signal path and the normal end of training.

Tested with a real `scancel` of a 1-GPU run on the scratch recipe: it wrote `checkpoint_00000186.pt` at the step it was cancelled on (checkpoint frequency was set to 100k, so this can only come from the stop path). The new unit test fails without the change.


training/train.py
```diff
@@ -170,28 +170,27 @@ def main(config_path: str):
...
+        # across restarts, so a fixed seed would replay the same
+        # permutation from the top and bias coverage toward its head.
+        seed=args.seed + start_step,
+        shuffle=args.data.shuffle,
+        num_procs=args.data.loader_procs,
+        num_bucket_batches=args.data.num_bucket_batches,
+        prompt_trim_max_sec=args.data.prompt_trim_max_sec,
+        final_punct_dropout=args.data.final_punct_dropout,
     )
+    train_loader = iter(train_data)
```
manukyutai: Why this change?

manukyutai (approved):

PR #324, opened 2026-09-23
Default every shipped model to temperature 0.3

manu-lm:
Fixes #322.

Configs that don't set `default_temperature` (every non-English model and `english_2026-01`) fell back to the schema default of 0.7, while the English models ship at 0.3 (#223). This makes 0.3 the schema default, so every shipped config samples at 0.3; a config can still set its own value.

## Why 0.3

Cross-sentence eval per language (CML-TTS test speakers; MLS for French), cfg 1.0, whisper-large-v3 WER / UTMOS:

| model | temp 0.3 | temp 0.7 (today) |
|---|---|---|
| french_24l (2026-07) | 5.57% / 3.00 | 5.61% / 2.90 |
| french (6L) | 4.64% / 3.00 | 5.07% / 2.85 |
| french_24l | 4.70% / 2.97 | 5.28% / 2.80 |
| spanish_24l | 4.97% / 3.21 | 5.46% / 3.15 |
| italian_24l | 9.09% / 3.31 | 20.65% / 3.23 |
| portuguese_24l | 10.66% / 2.79 | 12.85% / 2.82 |
| german_24l | 6.46% / 3.40 | 6.80% / 3.35 |

Speaker similarity is within ±0.006 across the two temperatures for every model.

## Changes

- `Config.default_temperature` and `DEFAULT_TEMPERATURE`: 0.7 -> 0.3.
- CLI help, `load_model` docstring and the docs drop "0.7 otherwise".
- Test: every shipped config resolves to 0.3.

PR #325, opened 2026-09-23
French: rewrite characters the models never saw in training (quotes, curly apostrophes)

manu-lm:
The French models from #321 speak filler syllables where the text contains characters their training text never had. MLS and Emilia-FR contain no straight or curly quotes, no curly apostrophes, almost no colons and no brackets, so those tokens' embeddings were never trained. Curly apostrophes are what phones and macOS type by default.

This adds `replace_characters`, a per-model map applied in `prepare_text_prompt` before tokenization (`""` deletes), and sets it for `french` and `french_24l`. The default is empty, so no other model changes.

## Measurements

1127 LibriSpeech-PC test-clean sentences translated to French with localized names, 18 MLS dev speakers as voice prompts, cfg 1.0, temp 0.3, whisper-large-v3 WER:

| input text | french (6L) | french_24l | french_24l before #321 |
|---|---|---|---|
| as written (« » quotes) | 4.93% | 5.02% | 5.59% |
| straight `"` quotes | 8.43% | 7.46% | 6.05% |
| curly `’` apostrophes | 12.17% | 12.58% | - |
| with this map | **3.84%** | **4.02%** | 6.51% |

The previous french_24l did learn these characters (mapping them makes it worse), which is why the map is set per model.

The retrained Spanish, Italian, Portuguese and German models under way have the same gaps (CML-TTS has no quotes, curly apostrophes or brackets). They will set their own map when they ship.

The second commit is the same `uv.lock` re-resolution as #324 (7-day exclude-newer window), needed for pre-commit to pass.


uv.lock
manukyutai: Needed?

manu-lm: Not here: dropped it, the re-resolution lands with #324, so this one goes in after #324.

pocket_tts/config/french.yaml
```diff
@@ -2,6 +2,23 @@ weights_path: hf://kyutai/pocket-tts/languages/french/model.safetensors@d0ebc5de
 weights_path_without_voice_cloning: hf://kyutai/pocket-tts-without-voice-cloning/languages/french/model.safetensors@8843db76457a91db32077edf8dfcd1c0e3e755fd
 
 remove_semicolons: true
+# Characters the French training text (MLS + Emilia-FR) never contains: their tokens are untrained
```
manukyutai: Simplify comment:
" replace characters which were absent of the training data" or similar wording. No need to give numbers

manukyutai (approved):

